PR #36, opened 2021-04-29
NCS-143 Create page not found screen

LewisJohnEvans:
Creation of page not found stop screen redirected from any unrecognised urls

### DIFF
--- a/src/controllers/error.controller.ts
+++ b/src/controllers/error.controller.ts
@@ -1,0 +1,8 @@
+import { Request, Response } from "express";
+import { Templates } from "../types/template.paths";
+
+const pageNotFound = (req: Request, res: Response) => {
+  return res.render(Templates.ERROR_404);
+};
+
+export default pageNotFound;
--- a/src/controllers/error.controller.ts
+++ b/src/controllers/error.controller.ts
@@ -2,7 +2,7 @@ import { Request, Response } from "express";
 import { Templates } from "../types/template.paths";
 
 const pageNotFound = (req: Request, res: Response) => {
-  return res.render(Templates.ERROR_404);
+  return res.status(404).render(Templates.ERROR_404);
 };
 
 export default pageNotFound;
--- a/test/controllers/error.controller.unit.ts
+++ b/test/controllers/error.controller.unit.ts
@@ -1,0 +1,17 @@
+jest.mock("ioredis");
+
+import mocks from "../mocks/all.middleware.mock";
+import request from "supertest";
+import app from "../../src/app";
+
+const EXPECTED_TEXT = "Page not found - File a confirmation statement";
+const INCORRECT_URL = "/confirmation-statement/company-numberr";
+
+describe("error controller test", () => {
+  it("should return page not found screen if page url is not recognised", async () => {
+    const response = await request(app)
+      .get(INCORRECT_URL);
+    expect(response.text).toContain(EXPECTED_TEXT);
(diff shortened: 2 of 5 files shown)


test/controllers/error.controller.unit.ts
```diff
@@ -0,0 +1,17 @@
...
+import app from "../../src/app";
+
+const EXPECTED_TEXT = "Page not found - File a confirmation statement";
+const INCORRECT_URL = "/confirmation-statement/company-numberr";
+
+describe("error controller test", () => {
+  it("should return page not found screen if page url is not recognised", async () => {
+    const response = await request(app)
+      .get(INCORRECT_URL);
+    expect(response.text).toContain(EXPECTED_TEXT);
```
ICoxCH, line 14: You should also expect the response.status to equal 404.
I think you'll need to modify the error controller to actually do this

src/controllers/error.controller.ts
```diff
@@ -0,0 +1,8 @@
+import { Request, Response } from "express";
+import { Templates } from "../types/template.paths";
+
+const pageNotFound = (req: Request, res: Response) => {
+  return res.status(404).render(Templates.ERROR_404);
+};
+
+export default pageNotFound;
```
ICoxCH, line 8: just thinking ahead to when we will need to also have an "500 Something went wrong" page.
It may be best not to have a default export on this file.

Export the pageNotFound function like so: `export const pageNotFound = ...`
And on app.ts import it like `import { pageNotFound } from "./controllers/error.controller";`
And use it like `app.use(pageNotFound);`

This way we can just add an "internalServerError" function to the error controller class without breaking anything

LewisJohnEvans: I can make that change if you think it's best, i used:
https://github.com/companieshouse/strike-off-objections-web/blob/master/src/controllers/error.controller.ts
As a basis and in that file it looked like the export could be updated later on to include other errors. Unless i'm miss understanding how that works.

ICoxCH: Ah yes I see how that works now, in that case leave it be this will work well, i'll approve

ICoxCH (approved):

views/page-not-found.html
```diff
@@ -0,0 +1,20 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Page not found - File a confirmation statement
+{% endblock %}
+{% block signOut %}{% endblock %}
```
markpit, line 6: a bit late I know, but line 6 can be removed - the signOut block we aren't using in out layout.html so we don't need to override it with an empty block. We may add it in future but don't seem to be using it yet.

views/page-not-found.html
```diff
@@ -0,0 +1,20 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">
+        Page not found
+      </h1>
+      <p>If you typed the web address, check it is spelled correctly.</p>
+      <p>If you pasted the web address, check you copied the entire address.</p>
+      <p>If the web address is correct or you selected a link, the page may have moved.
+        <a href="/confirmation-statement">Go to the File a confirmation statement start page</a>.
+
+      </div>
```
markpit, line 18: bottom divs and endblock out of alignment

LewisJohnEvans: Hi mark, i've pushed up the changes here and for the above but due to the PR already being merged i've had to make a new PR to get the change merged
new PR: https://github.com/companieshouse/confirmation-statement-web/pull/38

PR #39, opened 2021-04-29
Add accessibility statement page

kheseyongya:
JIRA TICKET - [NCS-163](https://companieshouse.atlassian.net/browse/NCS-163) 
Added an accessbility statement page, whitelist function in authentication middleware and unit tests.

### DIFF
--- a/test/middleware/authentication.middleware.unit.ts
+++ b/test/middleware/authentication.middleware.unit.ts
@@ -24,6 +24,10 @@ const expectedAuthMiddlewareConfig: AuthOptions = {
 
 describe("authentication middleware tests", () => {
 
+  beforeEach(() => {
+    mockAuthMiddleware.mockClear();
+  });
+
   it("should call CH authentication library", () => {
     authenticationMiddleware(req, res, next);
 
@@ -31,11 +35,11 @@ describe("authentication middleware tests", () => {
     expect(mockAuthReturnedFunction).toHaveBeenCalledWith(req, res, next);
   });
 
-  it("should not call authentication middleware when loading accessibility page", () => {
+  it("should not redirect to sign in when url is on whitelist", () => {
     req.originalUrl = CONFIRMATION_STATEMENT + ACCESSIBILITY_STATEMENT;
     authenticationMiddleware(req, res, next);
 
-    expect(mockAuthMiddleware).toBeCalled;
+    expect(mockAuthMiddleware).not.toHaveBeenCalled();
   });
 
 });
--- a/test/middleware/authentication.middleware.unit.ts
+++ b/test/middleware/authentication.middleware.unit.ts
@@ -3,6 +3,7 @@ jest.mock("@companieshouse/web-security-node");
 import { authMiddleware, AuthOptions } from "@companieshouse/web-security-node";
 import { NextFunction, Request, Response } from "express";
 import { authenticationMiddleware } from "../../src/middleware/authentication.middleware";
+import { ACCESSIBILITY_STATEMENT, CONFIRMATION_STATEMENT } from "../../src/types/page.urls";
 
 // get handle on mocked function and create mock function to be returned from calling authMiddleware
 const mockAuthMiddleware = authMiddleware as jest.Mock;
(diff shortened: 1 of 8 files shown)


markpit:
need to add the jira number to the commit title (you can click edit at top of page in github)

test/middleware/authentication.middleware.unit.ts
```diff
@@ -29,4 +30,12 @@ describe("authentication middleware tests", () => {
     expect(mockAuthMiddleware).toHaveBeenCalledWith(expectedAuthMiddlewareConfig);
     expect(mockAuthReturnedFunction).toHaveBeenCalledWith(req, res, next);
   });
+
+  it("should not call authentication middleware when loading accessibility page", () => {
```
ICoxCH, line 34: Maybe call this test "should not redirect to sign in when url is on whitelist"

test/middleware/authentication.middleware.unit.ts
```diff
@@ -29,4 +30,12 @@ describe("authentication middleware tests", () => {
...
+    req.originalUrl = CONFIRMATION_STATEMENT + ACCESSIBILITY_STATEMENT;
+    authenticationMiddleware(req, res, next);
+
+    expect(mockAuthMiddleware).toBeCalled;
```
ICoxCH, line 38: this expect isn't really doing anything.

I would set the test up similarly to the test in test/routes/routes.spec.unit.ts

But expect the response.text not to contain "sign in"

That way the test is asserting that if the url is whitelisted, it does not redirect them to the sign in page

markpit, line 38: i think here you'd want to expect authMiddleware not to be called. What might be happening is that the mockAuthMiddleware is still remembering calls from the previous test. What we should do is add a beforeEach(()=> {}). function after the describe.
In there we can say mockAuthMiddleware.mockClear()
also I would clear the req.originalUrl between tests, so something like req.originalUrl = ""

views/accessibility-statement.html
```diff
@@ -0,0 +1,85 @@
...
+          <li>are developing
+            <a href="https://gds.blog.gov.uk/2018/06/20/creating-the-uk-governments-accessibility-empathy-lab/">an assistive technology lab</a>
+            in our Cardiff office to improve the testing of our services and raise awareness of accessibility</li>
+
+          <p>
+            <a href="https://companieshouse.blog.gov.uk/category/user-research/">Read our blogs about usability</a>
+            to find out more about our user research work.</p>
+          <p>This statement was prepared on 22 April 2021. It was last updated on 22 April 2021.</p>
+
+        </div>
```
markpit, line 83: the bottom 3 lines seem a bit out of alignment

kheseyongya: Amended

test/middleware/authentication.middleware.unit.ts
kheseyongya: Amended

test/middleware/authentication.middleware.unit.ts
kheseyongya: I have included those function and can confirm the unit test is passing. Thanks

test/middleware/authentication.middleware.unit.ts
```diff
@@ -24,18 +24,24 @@ const expectedAuthMiddlewareConfig: AuthOptions = {
 
 describe("authentication middleware tests", () => {
 
+  beforeEach(() => {
+    mockAuthMiddleware.mockClear();
+    next.mockClear();
```
markpit, line 29: i would reset the req.originalUrl back to it's original value here so that if we add more tests in the future the url won't still be set to CONFIRMATION_STATEMENT + ACCESSIBILITY_STATEMENT. 
I would set it back to URL ----   req.originalUrl = URL.   The req will be shared between tests as it is declared in the global scope before the tests start.

kheseyongya: Makes sense. That change should be in now.

markpit (approved):

PR #41, opened 2021-04-29
ncs-59 Create confirm company page.

AlfredBwomezi:
Create confirm company page.

### DIFF
--- a/src/types/page.urls.ts
+++ b/src/types/page.urls.ts
@@ -1,3 +1,10 @@
+import { Templates } from "./template.paths";
+
+const SEPARATOR = "/";
+
 export const CONFIRMATION_STATEMENT = "/confirmation-statement";
 export const COMPANY_NUMBER = "/company-number";
-export const COMPANY_LOOKUP = "/company-lookup/search?forward=/confirmation-statement/confirm-company?companyNumber={companyNumber}"
+export const COMPANY_LOOKUP = "/company-lookup/search?forward=/confirmation-statement/confirm-company?companyNumber={companyNumber}";
+export const CONFIRM_COMPANY = SEPARATOR + Templates.CONFIRM_COMPANY;
+
+export const CONFIRM_COMPANY_PATH = CONFIRMATION_STATEMENT + CONFIRM_COMPANY;
--- a/src/types/page.urls.ts
+++ b/src/types/page.urls.ts
@@ -2,7 +2,7 @@ import { Templates } from "./template.paths";
 
 const SEPARATOR = "/";
 
-export const ACCESSIBILITY_STATEMENT = "/accessibility-statement";
+export const ACCESSIBILITY_STATEMENT = SEPARATOR + Templates.ACCESSIBILITY_STATEMENT;
 export const CONFIRM_COMPANY = SEPARATOR + Templates.CONFIRM_COMPANY;
 export const CONFIRMATION_STATEMENT = "/confirmation-statement";
 export const COMPANY_NUMBER = "/company-number";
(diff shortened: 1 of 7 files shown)


src/types/page.urls.ts
```diff
@@ -1,4 +1,11 @@
+import { Templates } from "./template.paths";
+
+const SEPARATOR = "/";
+
 export const ACCESSIBILITY_STATEMENT = "/accessibility-statement";
+export const CONFIRM_COMPANY = SEPARATOR + Templates.CONFIRM_COMPANY;
```
markpit, line 6: you could do the same with the accessibility-statement, change it to use the template name as the page name is the same as template.

test/controllers/confirm.company.controller.unit.ts
```diff
@@ -0,0 +1,32 @@
+jest.mock("ioredis");
+jest.mock("../../src/middleware/authentication.middleware");
```
markpit, line 2: see company.number.controller.unit.ts, if you add
import mocks from "../mocks/all.middleware.mock";

it should mean you can remove all the middleware mocking (as I moved it all into the test/mocks/ files.

markpit: but you may have to do at least one expect on the const you import to check that one of the middlewares were called, otherwise it might complain that the import is unused.

views/confirm-company.html
```diff
@@ -0,0 +1,133 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Confirm the company
+{% endblock %}
+{% block signOut %}{% endblock %}
```
markpit, line 6: we can remove the sign out block. We removed it from the layout so we don't need to override it in this file.

views/confirm-company.html
```diff
@@ -0,0 +1,133 @@
...
+
+        </p>
+        {% endset %}
+        {% set coTypeWarningHTML %}
+        This company does not meet all the criteria, you will need to file a confirmation statement using our
+        <a href="https://ewf.companieshouse.gov.uk//seclogin?tc=1">WebFiling service</a>.<br><br>
+        This service can only be used to file confirmation statements for private limited companies of a certain size.
+
+        {% endset %}
+        {{ govukTable({
```
markpit, line 39: indentation will need fixing from here down to end of file

views/confirm-company.html
```diff
@@ -0,0 +1,131 @@
...
+              ]
+            ]
+          }) 
+        }}
+        {% if company.csDue === "notdue" %}
+          {{ govukWarningText({
+            html: notDueWarningHTML,
+            iconFallbackText: "Warning"
+          }) }}
+        {% else %}
```
markpit, line 105: looks like this empty else can be removed and same on line 112

test/controllers/confirm.company.controller.unit.ts
```diff
@@ -0,0 +1,17 @@
+jest.mock("ioredis");
```
markpit, line 1: should be able to remove the ioredis now as it will be imported from the session middleware mock (included in all.middleware.mocks)

AlfredBwomezi: When removed (the test still passes, but) you get the warning
 `A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.`

How about moving it to the middleware mock class?

markpit: ok, you can leave it and i'll investigate. it should already be moved to the mock middleware, i can take a look after merged.

markpit (approved):

PR #47, opened 2021-05-04
NCS-90 getting company profile data

markpit:
Updated company.profile service and it's test. Fixed confirm.company.controller test as well.

Had to add an extra guard clause in to check if the sdk response was undefined or null, otherwise we wouldn't be able to get the http status code, eg. if we got a 404 then we wouldn't be able to get the http status from the response as there would be no response. So I guess the http status might only be useful in rare circumstances but it should cover our bases by checking it.

Got the unit test to mock the api client. Think we we mocking the class that got returned instead of the function. Also managed to get the 'anonymous' object to be returned from the mock. Think we were trying to use mockResolvedValue when we should have used mockReturnValue (createApiClient does not return a promise)


test/services/company.profile.service.unit.ts
```diff
@@ -0,0 +1,138 @@
...
+  resource: {
+    accounts: {
+      nextAccounts: {
+        periodEndOn: "2019-10-10",
+        periodStartOn: "2019-01-01",
+      },
+      nextDue: "2020-05-31",
+      overdue: false,
+    },
+    companyName: "Girl's school trust",
```
ICoxCH, line 109: should we change this data to not be an old copy of live data. 
Maybe change the company number to 12345678, and the company name to "Test Company"

markpit: new version pushed

ICoxCH (approved):

PR #52, opened 2021-05-12
Ncs 192 confirm company post

mattch1:
Groundwork for private sdk call


src/services/confirmation.statement.service.ts
```diff
@@ -0,0 +1,33 @@
...
+import Resource from "@companieshouse/api-sdk-node/dist/services/resource";
+
+export const createConfirmationStatement = async (session: Session,
+                                                  transactionId: string): Promise<Resource<any>> => {
+    const client = createOAuthApiClient(session);
+    const csService: ConfirmationStatementService = client.confirmationStatementService;
+    return await csService.postNewConfirmationStatement(transactionId);
+}
+
+export const createOAuthApiClient = (session: Session): PrivateApiClient => {
```
ICoxCH, line 19: I think it would make more sense for the creation of the api client to be extracted into its own class, in case other services wish to use it.

src/controllers/confirm.company.controller.ts
```diff
@@ -1,11 +1,24 @@
-import { Request, Response } from "express";
+import {NextFunction, Request, Response} from "express";
```
ICoxCH, line 1: Have you linted this pr?
I think there should be spacing before and after the curlies

src/services/confirmation.statement.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    const signInInfo = session.data?.[SessionKey.SignInInfo];
+    if(signInInfo) {
+        const oAuth: string = signInInfo?.[SignInInfoKeys.AccessToken]?.[AccessTokenKeys.AccessToken] as string;
+        if (oAuth) {
+            return createPrivateApiClient(CHS_API_KEY, oAuth, INTERNAL_API_URL);
+        }
+    }
+    return logAndThrowError(new Error ("Error getting session keys"));
```
ICoxCH, line 27: Possibly be more specific about the error?
"Error retrieving session keys for creating api client"

src/services/confirmation.statement.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    const csService: ConfirmationStatementService = client.confirmationStatementService;
+    return await csService.postNewConfirmationStatement(transactionId);
+}
+
+export const createOAuthApiClient = (session: Session): PrivateApiClient => {
+    const signInInfo = session.data?.[SessionKey.SignInInfo];
+    if(signInInfo) {
+        const oAuth: string = signInInfo?.[SignInInfoKeys.AccessToken]?.[AccessTokenKeys.AccessToken] as string;
+        if (oAuth) {
+            return createPrivateApiClient(CHS_API_KEY, oAuth, INTERNAL_API_URL);
```
ICoxCH, line 24: https://github.com/companieshouse/private-api-sdk-node/blob/master/src/index.ts#L13
We only want either the CHS_API_KEY or the users OAuth token.
As this method is called createOAuth client, I wouldn't pass in the ApiKey.

src/services/confirmation.statement.service.ts
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+    return logAndThrowError(new Error ("Error getting session keys"));
+};
+
+const logAndThrowError = (error: Error) => {
```
ICoxCH, line 30: I think we have an incredibly similar method elsewhere in the project.
Perhaps move that and this one to the logger util class and export it.

src/controllers/confirm.company.controller.ts
markpit: sounds like your githook is not activated, run
`find .githooks -type f -exec ln -sf ../../{} .git/hooks/ \;`

src/utils/properties.ts
```diff
@@ -26,3 +26,7 @@ export const CACHE_SERVER = getEnvironmentVariable("CACHE_SERVER");
 export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentVariable("SHOW_SERVICE_OFFLINE_PAGE");
 
 export const CHS_API_KEY = getEnvironmentVariable("CHS_API_KEY");
+
+export const INTERNAL_API_URL = getEnvironmentVariable("INTERNAL_API_URL");
+
+export const PRIVATE_SDK_FEATURE_FLAG = getEnvironmentVariable("PRIVATE_SDK_FEATURE_FLAG");
```
markpit, line 32: Steve wanted us to follow the CH style guide for feature flag naming - https://github.com/companieshouse/styleguides/blob/master/feature_flag.md

src/services/confirmation.statement.service.ts
```diff
@@ -0,0 +1,33 @@
...
+import PrivateApiClient from "private-api-sdk-node/dist/client";
+import {CHS_API_KEY, INTERNAL_API_URL} from "../utils/properties";
+import {ConfirmationStatementService} from "private-api-sdk-node/dist/services/confirmation-statement";
+import Resource from "@companieshouse/api-sdk-node/dist/services/resource";
+
+export const createConfirmationStatement = async (session: Session,
+                                                  transactionId: string): Promise<Resource<any>> => {
+    const client = createOAuthApiClient(session);
+    const csService: ConfirmationStatementService = client.confirmationStatementService;
+    return await csService.postNewConfirmationStatement(transactionId);
```
markpit, line 16: the csService.postNewConfirmationStatement is missing a return type on the function. It returns Resource<ConfirmationStatementSubmission> so this createConfirmationStatement function should return Promise<Resource<ConfirmationStatementSubmission>> instead of Promise<Resource<any>>

test/controllers/confirm.company.controller.unit.ts
```diff
@@ -8,6 +10,7 @@ import { getCompanyProfile } from "../../src/services/company.profile.service";
 import { validCompanyProfile } from "../mocks/company.profile.mock";
 
 const mockGetCompanyProfile = getCompanyProfile as jest.Mock;
+const mockCreateConfirmationStatement = createConfirmationStatement as jest.Mock;
```
markpit, line 13: you should add the mock to the beforeEach to clear/reset it

test/services/api/api.service.unit.ts
```diff
@@ -0,0 +1,9 @@
+import { createOAuthApiClient } from "../../../src/services/api/api.service";
+import { getSessionRequest } from "../../mocks/session.mock";
+
+describe ("", () => {
```
markpit, line 4: describe needs a description

test/services/confirmation.statement.service.unit.ts
```diff
@@ -0,0 +1,33 @@
...
+const mockPostNewConfirmationStatement
+    = ConfirmationStatementService.prototype.postNewConfirmationStatement as jest.Mock;
+const mockCreatePrivateApiClient = createPrivateApiClient as jest.Mock;
+
+const transactionId = "12345";
+
+describe ("Confirmation statement api service unit tests", () => {
+
+  beforeEach (() => {
+    mockCreatePrivateApiClient.mockReturnValue({
```
markpit, line 21: i'd add the mockPostNewConfirmationStatement in beforeEach as well to clear/reset

test/services/confirmation.statement.service.unit.ts
```diff
@@ -0,0 +1,33 @@
...
+      confirmationStatementService: ConfirmationStatementService.prototype
+    } as PrivateApiClient
+    );
+  });
+
+  it ("should create confirmation statement is called in the private sdk", async () => {
```
markpit, line 27: "should create confirmation statement is called in the private sdk" I think is meant to say "should call create confirmation statement in the private sdk"

src/services/confirmation.statement.service.ts
mattch1: That's the othe rmethod in the  cs privatesdk class "postConfirmationStatementSubmission" you're thinking of.

markpit: that's the one - https://github.com/companieshouse/private-api-sdk-node/blob/master/src/services/confirmation-statement/service.ts#L33
The return type needs to be added to the private sdk on the function https://github.com/companieshouse/private-api-sdk-node/blob/master/src/services/confirmation-statement/service.ts#L33. Then we can return a `Resource<ConfirmationStatementSubmission>` from the function you are adding now instead of `any`

markpit: right i think i am looking at the wrong one

mattch1: You are, but that's a good point about the missing return type for submission but that'll need to be changed in private sdk node.

test/services/api/api.service.unit.ts
```diff
@@ -0,0 +1,9 @@
+import { createOAuthApiClient } from "../../../src/services/api/api.service";
+import { getSessionRequest } from "../../mocks/session.mock";
+
+describe ("Test node session handler authorization for private sdk", () => {
```
markpit, line 4: can you add tests to test the error gets thrown? need to cover session.data being undefined, signInInfo being undefined and AccessToken being undefined. We had similar issues with date formatter where we didn't test all paths through the function and that caused the bug.

src/services/confirmation.statement.service.ts
markpit: ok, we can look at that separately

test/services/api/api.service.unit.ts
mattch1: Ok added test now passing.

test/services/api/api.service.unit.ts
```diff
@@ -0,0 +1,39 @@
...
+    } catch (error) {
+      expect(error.message).toBe(ERROR_MESSSAGE);
+    }
+  });
+
+  it("should throw error when no sign in info is present", () => {
+    try {
+      const session: Session = getEmptySessionRequest();
+      session.data = {};
+      createOAuthApiClient(session);
```
markpit, line 26: just needs a fail() after the call to the function (in each error test) so that the test won't pass if no error is thrown

markpit (approved):

PR #60, opened 2021-05-17
Fixing function style to arrow function in authentication.middleware

markpit:


src/middleware/authentication.middleware.ts
```diff
@@ -21,6 +21,6 @@ export const authenticationMiddleware = (req: Request, res: Response, next: Next
   return authMiddleware(authMiddlewareConfig)(req, res, next);
 };
 
-function isWhitelistedUrl(url: string): boolean {
+const isWhitelistedUrl = (url: string): boolean => {
```
ICoxCH, line 24: What's the difference between these two ways of doing a function?

kheseyongya (approved):

src/middleware/authentication.middleware.ts
markpit: i know they are newer, introduced as part of ES6, they can be shorter but apparently do have limitations if we wanted to use the this keyword. Just keeping in consistent with the others.

markpit: i thought arrow functions were a successor to function but they were introduced as more of a convenience for one liners and  anonymous functions - https://www.sitepoint.com/es6-arrow-functions-new-fat-concise-syntax-javascript/. After reading that we can actually use the arrow function in this instance as we can one line it. Not sure if we should be using it for our other functions now...

PR #65, opened 2021-05-19
Upgraded web-security-node and implemented company authentication

ICoxCH:


test/middleware/company.authentication.middleware.unit.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe("company authentication middleware tests", () => {
+
+  beforeEach(() => {
+    mockCompanyAuthMiddleware.mockClear();
+    next.mockClear();
+    req.originalUrl = URL;
+  });
+
+  it("should call CH authentication library when company pattern in url", () => {
```
mattch1, line 34: Might be worthwhile testing with different company number patterns so as to test that regex string in the middleware, for example "AB123456" and "ab123456"

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+  const authMiddlewareConfig: AuthOptions = {
+    chsWebUrl: CHS_URL,
+    returnUrl: req.originalUrl,
+    companyNumber: companyNumber
+  };
+
+  return authMiddleware(authMiddlewareConfig)(req, res, next);
+};
+
+const extractCompanyNumberFromRequest = (req: Request): string | undefined => {
```
markpit, line 26: you can get the url params from req.params, but you will need to set the app.use with the param name (like in spring)
something liek this:
app.use(urls.CONFIRMATION_STATEMENT + "/company/:companyNumber", companyAuthenticationMiddleware);

have a look at https://stackabuse.com/get-query-strings-and-parameters-in-express-js/ (the extracting route params section)

i haven't actually got it running myself but should be doable

ICoxCH: req.params isn't populated for middleware.
This is because its the router that populates that field, and the request hasn't been through the router for middleware such as this.

ICoxCH: the actual route handler will be able to access that though as the url is defined with a path paramater: https://github.com/companieshouse/confirmation-statement-web/pull/65/files#diff-b1c71f3697315abae06b8eaaf77b9f90eb1c41012510bd82b92b4a0dd2632941R13

markpit: ok

markpit (approved):

PR #69, opened 2021-05-19
NCS-122 Updating company authentication to use url param and extrac…

markpit:
…ted company number validation

Changed company.authentication.middleware to get number from req.params
Extracted teh company number validation into it's own module

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -11,7 +11,6 @@ import { sessionMiddleware } from "./middleware/session.middleware";
 import cookieParser from "cookie-parser";
 import { logger } from "./utils/logger";
 import { companyAuthenticationMiddleware } from "./middleware/company.authentication.middleware";
-import { COMPANY_AUTH_PROTECTED_BASE } from "./types/page.urls";
 
 const app = express();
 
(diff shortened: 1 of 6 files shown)


test/middleware/company.authentication.middleware.unit.ts
```diff
@@ -27,40 +36,36 @@ describe("company authentication middleware tests", () => {
...
+    mockSessionMiddleware.mockClear();
+    mockServiceAvailabilityMiddleware.mockClear();
+    mockAuthenticationMiddleware.mockClear();
+    mockLoggerErrorRequest.mockClear();
   });
 
-  it("should call CH authentication library when company pattern in url", () => {
-    companyAuthenticationMiddleware(req, res, next);
+  it("should call CH authentication library when company pattern in url", async () => {
+    await request(app).get(URL);
```
markpit, line 46: i thought it would be better to test the param extraction by sending a 'real' url rather then call middleware directly

src/app.ts
```diff
@@ -42,8 +42,7 @@ app.use(serviceAvailabilityMiddleware);
 app.use(`${urls.CONFIRMATION_STATEMENT}*`, sessionMiddleware);
 const userAuthRegex = new RegExp("^" + urls.CONFIRMATION_STATEMENT + "/.+");
 app.use(userAuthRegex, authenticationMiddleware);
-const companyAuthRegex = new RegExp("^" + urls.CONFIRMATION_STATEMENT + "/company/.*/");
-app.use(companyAuthRegex, companyAuthenticationMiddleware);
+app.use(`${urls.CONFIRMATION_STATEMENT}/company/:companyNumber`, companyAuthenticationMiddleware);
```
ICoxCH, line 45: might be worth adding a wildcard to the end of the matcher. So it matches any url starting with this

markpit: it appears to match anything starting with that string from manual testing in the browser and i changed the unit test to check for it 
const URL = CONFIRMATION_STATEMENT + COMPANY_AUTH_PROTECTED_BASE.replace(":companyNumber", "12345678") + "/test";
That seems to work ok. I'll have to find some documentation but it looks like the string you give app.use is a starts with.
found some - https://expressjs.com/en/4x/api.html#app.use

ICoxCH (approved):

ICoxCH (approved):

ICoxCH (approved):

PR #72, opened 2021-05-20
Ncs 158 invalid company status

mattch1:
Added company status validation check to confirm company controller. 
Moved the click continue url from the template to the controller.
Tested invalid statuses.


src/controllers/confirm.company.controller.ts
```diff
@@ -17,10 +18,36 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
+    } else {
+      return res.render(Templates.INVALID_STATUS);
+    }
+  } catch (e) {
+    next(e);
+  }
+};
+
+const isStatusEligible = (company: CompanyProfile): boolean => {
+  const status: string = company.companyStatus;
```
ICoxCH, line 36: the web shouldn't have any knowledge on what companies are eligible or not, that should be entirely contained to the api.
Which unfortunately would mean for this story to be accomplished the web will first need to create a transaction against a company.

So the flow will need to be confirm and continue -> redirect to auth page -> call api to check eligibility (for this story it will post internalapi/transaction/id/confirmation-statement/ -> render approproate stop screen

mattch1: It sounds like this needs more discussion to me, we really should settle on a plan for this.

src/controllers/confirm.company.controller.ts
```diff
@@ -17,10 +18,36 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
+      return res.redirect(url);
+    } else {
+      return res.render(Templates.INVALID_STATUS);
+    }
+  } catch (e) {
+    next(e);
+  }
+};
+
+const isStatusEligible = (company: CompanyProfile): boolean => {
```
markpit, line 35: move the eligible check out of the controller into it's own file. There is a validators folder now so makes sense to have a company status validator in there. Although check with Ian about wether the checks should be done in the api.

src/controllers/confirm.company.controller.ts
```diff
@@ -17,10 +18,36 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
 };
 
 export const post = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const company: CompanyProfile = await getCompanyProfile(req.query.companyNumber as string);
+    if (isStatusEligible(company)) {
+      create(req);
```
markpit, line 24: create is an async function so you need to await it or we won't get the response back

src/controllers/confirm.company.controller.ts
mattch1: Flagged ticket pending discussion.

src/controllers/confirm.company.controller.ts
```diff
@@ -17,10 +18,36 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
+      const url = CONFIRMATION_STATEMENT_TRADING_STATUS.replace(":companyNumber", company.companyNumber);
```
markpit, line 25: An enum for url params has been added since you wrote this, you can replace :companyNumber with urlParams.PARAM_COMPANY_NUMBER from page.urls.ts

src/controllers/confirm.company.controller.ts
```diff
@@ -17,10 +18,36 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
+  let isValid: boolean = true;
+  invalidStatuses.forEach((invalid) => {
+    if (status === invalid) {
+      isValid = false;
+    }
+  });
+  return isValid;
+};
+
+const create = async (req: Request) => {
```
markpit, line 47: can you rename the create function, it's not clear what it's creating when read in the function that calls it.

mattch1: In the function createConfirmationStatement is called so I just put create so as not to clash with that.

mattch1:
Closing PR and rmoving branch as the requirements have changed.

PR #76, opened 2021-05-24
previous change removed required trailing slash from trading status p…

ICoxCH:
…age route

### DIFF
--- a/src/types/page.urls.ts
+++ b/src/types/page.urls.ts
@@ -12,5 +12,5 @@ export const CONFIRMATION_STATEMENT = "/confirmation-statement";
 export const COMPANY_NUMBER = "/company-number";
 export const COMPANY_LOOKUP = "/company-lookup/search?forward=/confirmation-statement/confirm-company?companyNumber={companyNumber}";
 export const CONFIRM_COMPANY_PATH = CONFIRMATION_STATEMENT + CONFIRM_COMPANY;
-export const COMPANY_AUTH_PROTECTED_BASE = `/company/:${urlParams.PARAM_COMPANY_NUMBER}`;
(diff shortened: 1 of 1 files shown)


markpit (approved):

src/types/page.urls.ts
```diff
@@ -12,5 +12,5 @@ export const CONFIRMATION_STATEMENT = "/confirmation-statement";
 export const COMPANY_NUMBER = "/company-number";
 export const COMPANY_LOOKUP = "/company-lookup/search?forward=/confirmation-statement/confirm-company?companyNumber={companyNumber}";
 export const CONFIRM_COMPANY_PATH = CONFIRMATION_STATEMENT + CONFIRM_COMPANY;
-export const COMPANY_AUTH_PROTECTED_BASE = `/company/:${urlParams.PARAM_COMPANY_NUMBER}`;
+export const COMPANY_AUTH_PROTECTED_BASE = `/company/:${urlParams.PARAM_COMPANY_NUMBER}/`;
```
markpit, line 15: actually, that means the company auth middleware won't get executed for /company/12345678

ICoxCH: Maybe I was a bit quick on the merge.

Is that an issue though?
There isn't a route for /company/number so wouldn't actually present any issue

markpit: i guess not in the grand scheme of things, but i think it would be better if we prefixed a '/' to the trading status const like we do for COMPANY_NUMBER.  That way, anything starting /company/:id would get authed.

markpit: i realised after i had pressed approve!

PR #90, opened 2021-05-26
add env variable for piwik start goal id

CHJSkinner:
add env variable for piwik start goal id

### DIFF
--- a/src/controllers/start.controller.ts
+++ b/src/controllers/start.controller.ts
@@ -1,6 +1,9 @@
 import { Request, Response } from "express";
+import { CHS_URL, PIWIK_START_GOAL_ID } from "../utils/properties";
 import { Templates } from "../types/template.paths";
 
 export const get = (req: Request, res: Response) => {
-  return res.render(Templates.START, { CHS_URL: process.env.CHS_URL, START_GOAL_ID: process.env.PIWIK_START_GOAL_ID, templateName: Templates.START });
+  return res.render(Templates.START, { CHS_URL,
+    PIWIK_START_GOAL_ID,
+    templateName: Templates.START });
 };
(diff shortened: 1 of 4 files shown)


src/controllers/start.controller.ts
```diff
@@ -2,5 +2,5 @@ import { Request, Response } from "express";
 import { Templates } from "../types/template.paths";
 
 export const get = (req: Request, res: Response) => {
-  return res.render(Templates.START, { CHS_URL: process.env.CHS_URL, templateName: Templates.START });
+  return res.render(Templates.START, { CHS_URL: process.env.CHS_URL, START_GOAL_ID: process.env.PIWIK_START_GOAL_ID, templateName: Templates.START });
```
markpit, line 5: can you separate the page options onto separate lines?
eg
`return res.render(Templates.START, {
    CHS_URL: process.env.CHS_URL,
    START_GOAL_ID: process.env.PIWIK_START_GOAL_ID,
    templateName: Templates.START });`

markpit: also, by using process.env we are bypassing the src/util/properties.ts file (if you look in there you'll see that the getEnvironmentVariable is using process.env and does some checks on it first.
so, you should be able to import 'PIWIK_START_GOAL_ID' from properties.ts and use it like 
START_GOAL_ID: PIWIK_START_GOAL_ID

and, (sorry), looks like the existing CHS_URL can be done the same way, import CHS_URL from properties and use
CHS_URL: CHS_URL
BUT! - if you have a variable and value of the same name you can use a shortcut and just use CHS_URL
eg
{
  CHS_URL,
  START_GOAL_ID: PIWIK_START_GOAL_ID,
  templateName: Templates.START }

PR #92, opened 2021-05-27
NCS-234 Matomo trading status page

mattch1:
Matomo event tracking.

### DIFF
--- a/views/check-trading-status.html
+++ b/views/check-trading-status.html
@@ -71,17 +71,26 @@
           items: [
             {
               value: "yes",
-              text: "Yes"
+              text: "Yes",
+              attributes: {
+                "data-event-id": "trading-status-correct-yes"
+              }
             },
             {
               value: "no",
-              text: "No"
+              text: "No",
+              attributes: {
+                "data-event-id": "trading-status-correct-no"
+              }
             }
           ]
         }) }}
 
         {{ govukButton({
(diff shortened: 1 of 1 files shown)


views/check-trading-status.html
```diff
@@ -71,17 +71,26 @@ <h1 class="govuk-heading-xl">Check the trading status of shares</h1>
...
+              }
             }
           ]
         }) }}
 
         {{ govukButton({
-           text: "Continue"
+           text: "Continue",
+           attributes: {
+             "data-event-id": "trading-status-continue"
```
SVarallo, line 92: We sure this is the right data event id? Usually for continue buttons it's just a generic continue rather than a page specific one

mattch1: Right, do you have an example?

markpit (approved):

views/check-trading-status.html
SVarallo: In other services yeah we don't usually have any data events on the continue buttons 🤔 not sure why they asked for one to be added to this, might be worth asking the analysts on this one

SVarallo: Or at least in strike off, the other team has made a point of removing them. While we still have a generic "continue button" data event in PTF

SVarallo (approved):

PR #93, opened 2021-05-27
Add stop screen for webfiling company types

kheseyongya:
[NCS-127](https://companieshouse.atlassian.net/browse/NCS-127)

Add stop screen for companies that cannot use CHS service but can file through webfiling.

### DIFF
--- a/src/controllers/confirm.company.controller.ts
+++ b/src/controllers/confirm.company.controller.ts
@@ -39,12 +39,11 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 };
 
 const displayEligibilityStopPage = (res: Response, eligibilityStatusCode: EligibilityStatusCode) => {
-  if (eligibilityStatusCode === EligibilityStatusCode.INVALID_COMPANY_STATUS) {
-    return res.render(Templates.INVALID_COMPANY_STATUS);
-  } else if (eligibilityStatusCode === EligibilityStatusCode.INVALID_COMPANY_TYPE_USE_WEB_FILING) {
-    return res.render(Templates.USE_WEBFILING);
+  const stopPage: string = stopPages[eligibilityStatusCode];
+  if (!stopPage) {
+    throw new Error(`Unknown eligibilityStatusCode ${eligibilityStatusCode}`);
   }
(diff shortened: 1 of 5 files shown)


src/controllers/confirm.company.controller.ts
```diff
@@ -41,6 +41,8 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 const displayEligibilityStopPage = (res: Response, eligibilityStatusCode: EligibilityStatusCode) => {
   if (eligibilityStatusCode === EligibilityStatusCode.INVALID_COMPANY_STATUS) {
     return res.render(Templates.INVALID_COMPANY_STATUS);
+  } else if (eligibilityStatusCode === EligibilityStatusCode.INVALID_COMPANY_TYPE_USE_WEB_FILING) {
```
markpit, line 44: now we're adding more pages we can create a 'mapping' object to save us having all the ifs.

you could add an object to hold the mapping like this:
```
const stopPages = {
  [EligibilityStatusCode.INVALID_COMPANY_STATUS]: Templates.INVALID_COMPANY_STATUS,
  [EligibilityStatusCode.INVALID_COMPANY_TYPE_USE_WEB_FILING]: Templates.USE_WEBFILING
};
```

Then you can change the displayEligibilityStopPage function to get the template from the stopPages object
```
const displayEligibilityStopPage = (res: Response, eligibilityStatusCode: EligibilityStatusCode) => {
  const stopPage: string = stopPages[eligibilityStatusCode];
  if (!stopPage) {
    throw new Error(`Unknown eligibilityStatusCode ${eligibilityStatusCode}`);
  }
  return res.render(stopPage);
};
```

markpit: also need to add a test to check the USE_WEBFILING stop screen gets returned

markpit:
looks good, just checking what the process is as the story has a web test attached to it. We may need to wait until the auto-test is in place before approval in case of accidental merge.

markpit:
ok, i can approve it but check with testers and merge it once the auto test subtask is done.

markpit (approved):

PR #97, opened 2021-06-01
Ncs 56 create task list page

mattch1:
Task list page.


ICoxCH:
As you haven't added or changed version of any dependencies, could you remove the package-lock from this pr?
That file should only be included if you have modified the version of a dependency or added a new one

src/controllers/task.list.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+import { initTaskList, TaskList } from "../types/task.list";
+import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
+import { getCompanyProfile } from "../services/company.profile.service";
+import { SessionKey } from "@companieshouse/node-session-handler/lib/session/keys/SessionKey";
+import { SignInInfoKeys } from "@companieshouse/node-session-handler/lib/session/keys/SignInInfoKeys";
+
+export const get = async (req: Request, res: Response) => {
+  const session: Session = req.session as Session;
+  const signInInfo = session.data[SessionKey.SignInInfo] as object;
+  const companyNumber = signInInfo[SignInInfoKeys.CompanyNumber] as string;
```
ICoxCH, line 14: instead of all this. Why not just use the company number in the url?

src/controllers/trading.status.controller.ts
```diff
@@ -1,6 +1,16 @@
...
+import { Session } from "@companieshouse/node-session-handler";
 
 export const get = (req: Request, res: Response) => {
   return res.render(Templates.TRADING_STATUS);
 };
+
+export const post = (req: Request, res: Response) => {
+  const session: Session = req.session as Session;
+  const signInInfo = session.data["signin_info"] as object;
+  const companyNumber = signInInfo["company_number"] as string;
```
ICoxCH, line 13: instead of all this. Why not just use the company number in the url?

src/controllers/task.list.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+  const company: CompanyProfile = await getCompanyProfile(companyNumber);
+  console.log("Console log " + JSON.stringify(company));
```
ICoxCH, line 16: this should use the logger not the console. (Or is it left over from debugging in which case remove)

src/controllers/task.list.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+  const taskList: TaskList = initTaskList();
+  session[ TASK_LIST ] = taskList;
```
ICoxCH, line 18: Why does the task list need to be stored in the session?
The story is to just render the task list template. We don't need to store anything in the session to achieve that

src/controllers/trading.status.controller.ts
mattch1: Was in two minds about this we have always taken it from the session before, but if people are happy for me to get this from the url it would be less code.

ICoxCH:
the task list template also looks as if it could do with some general cleaning up.
It's got duplicated html. And lots of logic im not sure we need at the moment. 
It also has prototype specific logic that should be removed. Such as line 317

src/controllers/task.list.controller.ts
markpit, line 18: We can't use the session to store the task list, it will have to be stored in mongo and updated/accessed via the api. This will allow the 'save and sign out' to work as the mongo submission will get updated as they go through the screens.

mattch1:
Not sure what should be kept in and what should be taken out of the html - will  start by splitting the file up and use the includes folder to make this more readable.

views/task-list.html
```diff
@@ -0,0 +1,331 @@
...
+                  <span class="app-task-list__task-name">
+                    {% if tradingStatus === '1' %}
+                      <a href="/confirmation-statement/active-pscs" aria-describedby="read-declaration-status">
+                        People with significant control (PSC)
+                      </a>
+                    </span>
+                    {% if taskList.pscStatement === 'yes' %}
+                      <strong class="govuk-tag app-task-list__tag" id="eligibility-status">
+                        Checked</strong>
+                    {% elseif taskList.additionalPscs === 'yes' or  taskList.pscs === 'no' or activePscs === 'no' %}
```
markpit, line 130: the multiple tasklist checks should be removed, we should only pass a flag in to the template to toggle the status of a task

markpit, line 130: the multiple tasklist checks should be removed, we should only pass a flag in to the template to toggle the status of a task.
basically we just want the html framework and styling of the page and the indicators that change (including the button disabled/enabled) should be passed in as flags/values from the controller. So the template will be told what to do by the controller. As we will be getting that info from teh api at a later date, it is ok to just pass in a dummy set of flags for this story.

src/types/task.list.ts
```diff
@@ -0,0 +1,52 @@
+import { DateTime } from "luxon";
+import { readableFormat } from "../utils/date.formatter";
+
+export interface TaskList {
```
markpit, line 4: A lot of these look they they come from the prototype. for our template we'll just need indicators/flags to toggle the status of each task. Looking at the story there only seems to be 5 or 6 tasks on the screen.

mattch1: 7 tasks for LTD companies and 4 for LLPs.

src/types/task.list.ts
```diff
@@ -0,0 +1,52 @@
...
+    email: string;
+    exemption: string;
+    activeOfficers: string;
+    activePscs: string;
+    activeMembers: string;
+    additionalPscs: string;
+    additionalOfficers: string;
+    additionalMembers: string;
+    members: string;
+    moment: string;
```
markpit, line 15: moment looks like the prototype was using the moment library to generate a date. seeing as we aren't using moment, rename it to something better like 'todaysDate'

src/controllers/task.list.controller.ts
```diff
@@ -0,0 +1,15 @@
+import { Request, Response } from "express";
+import { Templates } from "../types/template.paths";
+import { initTaskList, TaskList } from "../types/task.list";
+import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
+import { getCompanyProfile } from "../services/company.profile.service";
+
+export const get = async (req: Request, res: Response) => {
+  const companyNumber = req.params.companyNumber as string;
```
markpit, line 8: there is a urlParams.PARAM_COMPANY_NUMBER that can be used to identify the params name - req.params[urlParams.PARAM_COMPANY_NUMBER]

src/types/task.list.ts
markpit: you could create an enumerated type for the different states and pass that in to the template

mattch1: These really are just placeholders so that we can display the page there will be changes around this page as we develop this further.

views/task-list.html
mattch1: This might not be possible as we have:

 {% if taskList.sic === "yes" %}
            <strong class="govuk-tag app-task-list__tag" id="eligibility-status">
                Checked</strong>
        {% elseif taskList.sic === "no" %}
            <strong class="govuk-tag govuk-tag--blue app-task-list__tag" id="eligibility-status">In progress</strong>
        {% else %}
            <strong class="govuk-tag govuk-tag--grey app-task-list__tag" id="medical-information-status">Not checked</strong>
        {% endif %}

If we just have truthy or falsy then else and "Not Checked" will never be shown.

markpit: an enumerated type would solve it

markpit: just pass in the value of the enumerated type Checked/In progress/ Not checked

mattch1: I can go away and do something but it'd be good if we were on the same page can you give me an example of what form you think this should take, I'm presuming an enum with Checked,In progress, not checked

mattch1: Presumably we then replace:
completedTasks: string;
    email: string;
    exemption: string;
....

with

completedTasks: <enum_type>;
    email: <enum_type>;
    exemption:  <enum_type>;

src/controllers/task.list.controller.ts
mattch1: I'm sure that amounts to the same thing but will change this if it is preferred.

views/task-list.html
markpit: I was thinking of using an enum for TaskState. Each task has 3? states Checked/In progress/ Not checked. Then you'd have to use in the template {% if taskList.sicState === "Checked" %} then show 'checked' html

We might be able to define the html for Checked/In progress/ Not checked once each at top of template then insert the html for each depending on the state of the enum. So if we needed to show checked, we would use {{ set checkedHtml....  }.  to define the html then use {{checkedHtml}} to display it.  The html for the progress indicators seems to be the same
<strong class="govuk-tag govuk-tag--grey app-task-list__tag" id="medical-information-status">Not checked</strong>
but with different styling for each one.
Although not sure why they each have id="medical-information-status", you'd have to check with Oceanne if each should have a unique id and what they are used for.

the calculations (if a = yes and b = no etc) would get done before the page was rendered, so the template wouldn't have to work out if it should show checked/in progress/not checked, the template would just get told, "for sic task, show 'Checked'"

mattch1: I may not have understood this correctly, but I am unsure as to what we gain for the effort. I have just put up some refactored template code can we review this first?

src/controllers/task.list.controller.ts
markpit: it will be the same value but the benefit of using constants is that we only need to to change it in one place if it needs to change as opposed to searching out all occurrences of req.params.companyNumber.

src/types/task.list.ts
```diff
@@ -0,0 +1,56 @@
+import { DateTime } from "luxon";
+import { readableFormat } from "../utils/date.formatter";
+
+export enum Task {
```
markpit, line 4: The enum name doesn't look correct to me.  NOT_CHECKED, IN_PROGRESS, CHECKED are not types of Task, they are TaskState

src/controllers/task.list.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { initTaskList, TaskList } from "../types/task.list";
+import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
+import { getCompanyProfile } from "../services/company.profile.service";
+import { urlParams } from "../types/page.urls";
+
+export const get = async (req: Request, res: Response) => {
+  const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+  const company: CompanyProfile = await getCompanyProfile(companyNumber);
+  const taskList: TaskList = initTaskList();
+  const tradingStatus = "1";
```
markpit, line 12: it's not clear what tradingStatus 1 means, can you define a constant or enum for it?

mattch1: I wasn't sure either I am aware of code 0,1,3 on NCS-233 but 1 and 2 have been used here and I'm not sure which relates to which so that's why I put a placeholder I for now as the task is just to have  a page in place we can navigate to. The scope of this ticket does not feel all that clear.

mattch1:
Simplified this PR

src/types/task.list.ts
mattch1: Updated the name

src/controllers/trading.status.controller.ts
```diff
@@ -1,6 +1,13 @@
...
 import { Templates } from "../types/template.paths";
+import { TASK_LIST_PATH, urlParams } from "../types/page.urls";
 
 export const get = (req: Request, res: Response) => {
   return res.render(Templates.TRADING_STATUS);
 };
+
+export const post = (req: Request, res: Response) => {
+  const companyNumber =  req.params[urlParams.PARAM_COMPANY_NUMBER];
+  const url = TASK_LIST_PATH.replace(":companyNumber", companyNumber);
```
markpit, line 11: a minor thing, the ":companyNumber" can use the param constant eg
``:${urlParams.PARAM_COMPANY_NUMBER}``

markpit: > a minor thing, the ":companyNumber" can use the param constant eg
> `:${urlParams.PARAM_COMPANY_NUMBER}`

there are supposed to be 'backtick' quotes around my example but github interprets them as formatting, but you should be able to se what I mean.

markpit (approved):

PR #101, opened 2021-06-03
NCS 128 add stop screen for paper only filing companies

LewisJohnEvans:
NCS 128 add stop screen for paper only filing companies


src/controllers/confirm.company.controller.ts
```diff
@@ -24,26 +24,27 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
 };
 
-const displayEligibilityStopPage = (res: Response, eligibilityStatusCode: EligibilityStatusCode) => {
+const displayEligibilityStopPage = (res: Response, eligibilityStatusCode: EligibilityStatusCode, company: CompanyProfile) => {
   const stopPage: string = stopPages[eligibilityStatusCode];
   if (!stopPage) {
     throw new Error(`Unknown eligibilityStatusCode ${eligibilityStatusCode}`);
   }
-  return res.render(stopPage);
+  return res.render(stopPage, { company, templateName: Templates.USE_PAPER });
```
markpit, line 47: The templateName is specific to the USE_PAPER screen only, but so far we have 3 stop screens that will use this function and they'll all get the same templateName of USE_PAPER. The stopPage const will have the template name of the specific stop page so we can use that instead eg.
`return res.render(stopPage, { company, templateName: stopPage });`

markpit (approved):

PR #102, opened 2021-06-03
Ncs 114 company trading status

mattch1:
Naviagtion and validation for the trading status page.


src/controllers/trading.status.controller.ts
```diff
@@ -9,5 +10,15 @@ export const get = (req: Request, res: Response) => {
 export const post = (req: Request, res: Response) => {
   const companyNumber =  req.params[urlParams.PARAM_COMPANY_NUMBER];
   const url = TASK_LIST_PATH.replace(`:${urlParams.PARAM_COMPANY_NUMBER}`, companyNumber);
-  return res.redirect(url);
+  const tradingStatus: string = req.body.tradingStatus;
+  console.log("TRADING STATUS: " + tradingStatus);
+  if (tradingStatus === "yes") {
```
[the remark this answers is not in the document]
mattch1: Not here. If there were more I'd  probably use a switch.

views/trading-stop.html
```diff
@@ -0,0 +1,21 @@
+{% extends "layout.html" %}
+{% block pageTitle %}
+    Trading status not supported
+{% endblock %}
+{% block backLink %}
+    {{ govukBackLink({
+text: "Back",
+href: "/check-trading-status"
```
ICoxCH, line 8: is this back link correct? Should it not be prefixed with /company/{number}?
Which means we'll likely have to pass the company number into the template

mattch1: I don't know I have not really thought about backlinks I'd prefer to deal with the separately.

markpit: I don't think it's wise to move the story to ready for test if there is a known bug in the screen. You could fix in in a separate PR after this one, then the screen would be ready for test.

mattch1: It's not in the AC for the ticket and we are following the "little and often" approach I don't want to start bringing more stuff into the PRs as that is when they start to spriral out of control and become very large.

markpit: That's why I suggested fixing it in another PR, stories can have multiple PRs. It seems to me like we are adding in bugs that in all likelihood someone else will have to fix.

ICoxCH: Can you remove the back link entirely from this pr?
And in a seperate pr add the correct back link in?

ICoxCH (approved):

markpit (approved):

PR #103, opened 2021-06-03
NCS-165 filing not due warning message

markpit:
https://companieshouse.atlassian.net/browse/NCS-165
Adding a warning message if nextMadeUpTo date is in the future

Had to split the formatting of the company profile so we can use the dates before they are formatted which resulted in a few tests breaking and so had to update those.

Added a new date function to **date.formatter** and seeing as it isn't purely a formatting module now, i've renamed it to '**date**'


src/controllers/confirm.company.controller.ts
```diff
@@ -11,21 +11,50 @@ import {
...
+  }
+  return due;
 };
 
 export const post = async (req: Request, res: Response, next: NextFunction) => {
-  const session: Session = req.session as Session;
-  const companyNumber = req.query["companyNumber"] as string;
-  const company: CompanyProfile = await getCompanyProfile(req.query.companyNumber as string);
   try {
+    const session: Session = req.session as Session;
```
ICoxCH, line 55: is this session variable actually used?

markpit: looks like it is passed into the eligibility service and is used to create the OAuth API Client in there to make the api call.
It was already there but I noticed those 3 lines were outside the try {} so if they threw an error then page would hang. So moved them inside the try{}.

ICoxCH (approved):

PR #104, opened 2021-06-03
no filing required stop screen 

CHJSkinner:
no filing required stop screen


test/controllers/confirm.company.controller.unit.ts
```diff
@@ -104,16 +104,6 @@ describe("Confirm company controller tests", () => {
     expect(response.text).toContain("dissolved and struck off the register");
   });
 
-  it("Should redirect to error page when unrecognised eligibility code is returned", async () => {
```
ICoxCH, line 107: I'd argue this test still adds value.
Maybe change the mockResolvedValue to be some randome string like "abcdefgh"

markpit (approved):

test/controllers/confirm.company.controller.unit.ts
markpit: good point, ignore my approval,  you might need to 'cast' the random string to an EligibilityStatusCode using 'as'

CHJSkinner: Updated to string "abcdefg", test passes locally, no cast was needed

ICoxCH (approved):

PR #108, opened 2021-06-04
NCS-368 and NCS-369 all backlinks related to the trading status page

mattch1:
3 backlinks:
Back link from the trading status page and backlinks from both the yes and no paths to trading status.


src/controllers/task.list.controller.ts
```diff
@@ -7,9 +7,11 @@ import { urlParams } from "../types/page.urls";
 
 export const get = async (req: Request, res: Response) => {
   const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+  const back = `/confirmation-statement/company/${companyNumber}/trading-status`;
   const company: CompanyProfile = await getCompanyProfile(companyNumber);
   const taskList: TaskList = initTaskList();
   return res.render(Templates.TASK_LIST, {
+    back,
```
ICoxCH, line 14: This variable name on the templates may be better as something like "backLinkUrl" to better describe what it's doing.

ICoxCH (commented):
One small comment. And could you remove the package-lock please?

src/controllers/task.list.controller.ts
```diff
@@ -7,9 +7,11 @@ import { urlParams } from "../types/page.urls";
 
 export const get = async (req: Request, res: Response) => {
   const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+  const backLinkUrl = `/confirmation-statement/company/${companyNumber}/trading-status`;
```
markpit, line 10: can we use the PageUrls constants instead of adding the urls into the controller

views/trading-stop.html
```diff
@@ -3,6 +3,13 @@
     Trading status not supported
 {% endblock %}
 
+{% block backLink %}
+{{ govukBackLink({
```
markpit, line 7: indentation

markpit:
package-lock.json not needed

ICoxCH (approved):

PR #115, opened 2021-06-09
Ncs 267 company types display bug

mattch1:
In relation to the issues on the ticket:
Covered in unit and dev tests:

8 Converted/Closed – now returns Converted/Closed

30 Scottish Partnership – now returns Scottish Partnership 

26 United Kingdom Societas  - now returns United Kingdom Societas

Dev tested:

Uk-establishment – returns UK Establishment 

Not resolved:

36 United Kingdom Economic Interest Grouping - (ukeig)
19 Other type of company (in Northern Ireland)

The problem with these is down to the data we have in api-enumerations.

### DIFF
--- a/src/utils/api.enumerations.ts
+++ b/src/utils/api.enumerations.ts
@@ -1,5 +1,6 @@
 import * as fs from "fs";
 import * as yaml from "js-yaml";
+import { editCompanyTypeDisplay } from "./company.type.display.procressor";
 
 interface ApiEnumerationsConstants {
   [propName: string]: any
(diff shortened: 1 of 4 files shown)


src/utils/api.enumerations.ts
```diff
@@ -11,7 +12,8 @@ const apiConstants: ApiEnumerationsConstants = yaml.load(apiConstantsFile) as Ap
 export const lookupCompanyType = (companyTypeKey: string): string => {
   // we actually use the 'company_summary' values from the yaml file to
   //  display the company type (following the ch.gov.uk templates)
-  return apiConstants.company_summary[companyTypeKey] || companyTypeKey;
+  const displayValue: string = apiConstants.company_summary[companyTypeKey] || companyTypeKey;
+  return editCompanyTypeDisplay(companyTypeKey, displayValue);
```
markpit, line 16: we should update the api-enumerations to contain the missing values rather than try and calculate them. The api-enumerations are applied to all the other apps so the naming is consistent across the apps.
eg ukeig is a missing one, another team has added it to the api-enumerations develop branch (it should be merged into master).
I would get in touch with whoever added ukeig and see if it can be merged into master as there are probably others in develop that are not in master and would probably fix some of the missing values raised in the defect.
After that, any types that are missing should be added to api-enumerations and we can just read teh values from there. There is a confluence page for api-enumerations - https://companieshouse.atlassian.net/wiki/spaces/DEV/pages/126094319/api-enumerations

mattch1:
Closing this as the solution will require changes to the enumerations api submodule.

PR #125, opened 2021-06-14
update text on trading status page

ICoxCH:


mattch1:
Not sure I was told not include "of shares" we should check this with the analysts.

ICoxCH:
> Not sure I was told not include "of shares" we should check this with the analysts.

https://companieshouse.atlassian.net/browse/NCS-393

And lead analyst for the project has confirmed "of shares" is the correct wording

mattch1 (approved):

PR #132, opened 2021-06-15
NCS-267 Use company type

mattch1:
Discussed with analyst and decided that the content of company-type is what we require.


src/utils/api.enumerations.ts
```diff
@@ -9,9 +9,9 @@ const apiConstantsFile = fs.readFileSync("api-enumerations/constants.yml", "utf8
 const apiConstants: ApiEnumerationsConstants = yaml.load(apiConstantsFile) as ApiEnumerationsConstants;
 
 export const lookupCompanyType = (companyTypeKey: string): string => {
-  // we actually use the 'company_summary' values from the yaml file to
+  // we actually use the 'company_type' values from the yaml file to
```
markpit, line 12: i'd remove the comments completely now, they were only there to highlight that we were using company summary instead of type and function was called lookupCompanyType so comments not needed now.

markpit (approved):

PR #140, opened 2021-06-17
Update displayed filing date based on changed acceptance criteria

CHJSkinner:
Update displayed filing date based on changed acceptance criteria

### DIFF
--- a/src/controllers/task.list.controller.ts
+++ b/src/controllers/task.list.controller.ts
@@ -1,4 +1,4 @@
-import { NextFunction, Request, Response } from "express";
+import { Request, Response } from "express";
 import { Templates } from "../types/template.paths";
 import { initTaskList, TaskList } from "../types/task.list";
 import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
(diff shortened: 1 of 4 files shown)


src/controllers/task.list.controller.ts
```diff
@@ -1,28 +1,35 @@
...
-    company,
-    taskList
-  });
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+    const backLinkUrl = TRADING_STATUS_PATH.replace(`:${urlParams.PARAM_COMPANY_NUMBER}`, companyNumber);
+    const company: CompanyProfile = await getCompanyProfile(companyNumber);
+    if (!company.confirmationStatement) {
+      throw new Error("No confirmation Statement data in company profile");
```
markpit, line 15: i'd check with the analyst if we should show an error screen if the confirmation statement is missing. I think it might be missing from the company profile if it is the first year of the company and they haven't filed a CS before. In which case we might just want to show the current date? (analyst can advise)

test/controllers/task.list.controller.unit.ts
```diff
@@ -57,4 +57,14 @@ describe("Task list controller tests", () => {
     expect(response.text).toContain(expectedDate);
   });
 
+  it("Should show throw an error when confirmationStatent is missing", async () => {
+    validCompanyProfile.confirmationStatement = undefined;
+    mockGetCompanyProfile.mockResolvedValueOnce(validCompanyProfile);
+    mocks.mockAuthenticationMiddleware.mockClear();
+    const url = TASK_LIST_PATH.replace(":companyNumber", COMPANY_NUMBER);
+    const response = await request(app).get(url);
+    // expect(response.text).toContain(expectedDate);
```
markpit, line 66: commented out code can be removed

src/controllers/task.list.controller.ts
CHJSkinner: i just checked with an analyst and was told
> As soon as a company is incorporated, we generate the CS01 filing periods for them, and this info gets added to the Company_Profile delta then.
this means there will always be the dates needed, I will remove the check and error throw.
my confusion came from confirmationStatement being optional under the api-sdk-node, but from what ive been told that wouldnt ever be the case 🤷‍♂️

src/controllers/task.list.controller.ts
```diff
@@ -20,9 +20,9 @@ export const get = async (req: Request, res: Response) => {
 };
 
 const calculateFilingDate = (recordDate: string, companyProfile: CompanyProfile): string => {
-  const nextMadeUpToDate = companyProfile.confirmationStatement.nextMadeUpTo;
-  if (nextMadeUpToDate && isInFuture(nextMadeUpToDate)) {
+  const nextMadeUpToDate = companyProfile.confirmationStatement!.nextMadeUpTo;
```
markpit, line 23: i'd still have the checks for undefined in place, just in case.
You can use :
`const nextMadeUpToDate = companyProfile?.confirmationStatement?.nextMadeUpTo;`
which will return undefined if any of the question marked items are undefined and won't throw any errors.
This should be safe as isInFuture is safe handling undefined, it will return false if given an undefined (from looking at the unit test)

src/controllers/task.list.controller.ts
markpit: The get function should still be wrapped by a try - catch (as you had it before but without the if (!company.confirmationStatement)  check). This will catch any errors thrown from the api calls and show the error page. If we don't have it then the page will just hang if any errors occur. All the controller get and post functions should be wrapped in a try catch, this one must have slipped through the net previously.

src/controllers/task.list.controller.ts
CHJSkinner: Im not sure as if i remove the ! after confirmationstatement i get the expected
> Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

for the isInFuture method, it doesnt seem to be actually able to handle an undefined

CHJSkinner: Also in the case that somehow a company does have an undefined confirmation statement, it will direct them to service unavailable anyway which i believe is the correct procedure

CHJSkinner:
PR Moved to a new branch/pr as git wont let me commit anything -> https://github.com/companieshouse/confirmation-statement-web/pull/145

PR #153, opened 2021-06-24
Ncs 70 create nav psc page

mattch1:
Active pscs - navigation from the task list hyperlink and page with temporary hardcoded values.

I have added the hyperlink url to the task list model so that it can be passed to the template using the existing data we are already passing.


src/controllers/tasks/active.pscs.controller.ts
```diff
@@ -0,0 +1,9 @@
+import { Templates } from "../../types/template.paths";
+import { Request, Response } from "express";
+import { TASK_LIST_PATH, urlParams } from "../../types/page.urls";
+
+export const get = (req: Request, res: Response) => {
+  const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+  const backLinkUrl = TASK_LIST_PATH.replace(`:${urlParams.PARAM_COMPANY_NUMBER}`, companyNumber);
```
markpit, line 7: you can use the new function to build the url - https://github.com/companieshouse/confirmation-statement-web/blob/main/src/utils/url.ts#L3

src/types/task.list.ts
```diff
@@ -50,11 +52,11 @@ export interface TaskList {
   csDue: boolean;
 }
 
-export const initTaskList = (): TaskList => {
+export const initTaskList = (companyNumber: string): TaskList => {
   return {
     tasks: {
       officers: { isVisible: false, state: TaskState.NOT_CHECKED },
-      peopleSignificantControl: { isVisible: false, state: TaskState.NOT_CHECKED },
+      peopleSignificantControl: { isVisible: false, state: TaskState.NOT_CHECKED, url: addComapnyNumber(companyNumber, ACTIVE_PSCS_PATH) },
```
markpit, line 59: you can use that new url function here as well - https://github.com/companieshouse/confirmation-statement-web/blob/main/src/utils/url.ts#L3

views/tasks/active-pscs.html
```diff
@@ -0,0 +1,81 @@
...
+{% block backLink %}
+  {{ govukBackLink({
+    text: "Back",
+    href: backLinkUrl
+  }) }}
+{% endblock %}
+{% block content %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-three-quarters">
+            <form action="" method="post">
```
markpit, line 17: i think we can remove the empty action="" from the form and it should still work the same

src/types/task.list.ts
```diff
@@ -67,3 +69,7 @@ export const initTaskList = (): TaskList => {
     csDue: false
   };
 };
+
+const addComapnyNumber = (companyNumber: string, path: string) => {
```
SVarallo, line 73: typo

test/controllers/tasks/active.pscs.controller.unit.ts
```diff
@@ -0,0 +1,21 @@
...
+import { companyAuthenticationMiddleware } from "../../../src/middleware/company.authentication.middleware";
+
+const mockCompanyAuthenticationMiddleware = companyAuthenticationMiddleware as jest.Mock;
+mockCompanyAuthenticationMiddleware.mockImplementation((req, res, next) => next());
+
+const COMPANY_NUMBER = "12345678";
+
+describe("Active pscs controller tests", () => {
+  it("should navigate to the active pscs page", async () => {
+    mocks.mockAuthenticationMiddleware.mockClear();
```
markpit, line 16: just one more small thing, can you put the mockClear() in a beforeEach() function? that way it will be cleared when other tests get added in the future as well.

src/types/task.list.ts
mattch1: Updated this with a merge from main.

src/controllers/tasks/active.pscs.controller.ts
mattch1: Updated this with a merge from main.

src/types/task.list.ts
mattch1: Code no longer present.

views/tasks/active-pscs.html
mattch1: Removed

test/controllers/tasks/active.pscs.controller.unit.ts
mattch1: I'd probably add it in then as beforeEach implies that there are multiple tests there.

markpit: I'd say it is good practice to do it now as you aren't guaranteed to be the next person working on it and you might forget that it needs to be cleared between tests

markpit (approved):

PR #154, opened 2021-06-24
Ncs 62 create sic code page

kheseyongya:
[NCS-62
](https://companieshouse.atlassian.net/browse/NCS-62)
Create a sic code page and add paths to get to it.


src/controllers/confirm.sic.code.controller.ts
```diff
@@ -0,0 +1,15 @@
+import { NextFunction, Request, Response } from "express";
+import { TASK_LIST_PATH, urlParams } from "../types/page.urls";
+import { Templates } from "../types/template.paths";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+    const backLinkUrl = TASK_LIST_PATH.replace(`:${urlParams.PARAM_COMPANY_NUMBER}`, companyNumber);
```
AlfredBwomezi, line 8: Could consider using the [recently added function to populate the URL](https://github.com/companieshouse/confirmation-statement-web/pull/152/files). Could be something like
` const backLinkUrl = getUrlWithCompanyNumber(TASK_LIST_PATH, companyNumber);`

views/confirmation-statement/sic.html
```diff
@@ -0,0 +1,131 @@
...
+        {% set sicDetails %}
+        <p>The SIC (Standard Industrial Classification) code indicates the nature of the company’s business. A company can have up to 4 SIC codes.
+        </p>
+        <p>Check the
+          <a href="http://resources.companieshouse.gov.uk/sic/" target="_blank" rel="noopener noreferrer">condensed SIC codes list (opens in a new tab)</a>.</p>
+        {% endset %}
+
+        <h1 class="govuk-heading-l">Check the SIC codes</h1>
+        {{ govukDetails({
+  summaryText: "What is a SIC code?",
```
AlfredBwomezi, line 43: Alignment

views/confirmation-statement/sic.html
```diff
@@ -0,0 +1,131 @@
...
+          value: "no",
+          text: "No"
+        }
+      ]
+    }) }}
+        {% endif %}
+
+        {{ govukButton({
+text: "Continue"
+}) }}
```
AlfredBwomezi, line 127: Dito

src/types/template.paths.ts
```diff
@@ -4,6 +4,7 @@ export enum Templates {
     ERROR_404 = "page-not-found",
     SERVICE_OFFLINE = "service-offline",
     SERVICE_OFFLINE_MID_JOURNEY = "service-offline-mid-journey",
+    SIC = "confirmation-statement/sic",
```
markpit, line 7: I think Matt has created a subfolder called 'tasks' for the task list  pages, could you move yours into there? perhaps even create a sic subfolder under tasks if you think that might be clearer

views/includes/task-list/nature-of-business.html
```diff
@@ -4,7 +4,7 @@ <h2 class="app-task-list__section">
 <ul class="app-task-list__items">
     <li class="app-task-list__item">
         <span class="app-task-list__task-name">
-            <a href="/confirmation-statement/sic">
+            <a href="/confirmation-statement/company/{{company.companyNumber}}/sic">
```
markpit, line 7: could you build the url in the controller and pass it into the template as one of the templateOptions? this would make it easier to change the urls in future if we needed to as then we could just change the constants in pageUrls and we wouldn't have to change the urls in the templates.

views/includes/confirmation-statement-details/sic-details.html
```diff
@@ -0,0 +1,7 @@
+<p class="govuk-body-m">
```
markpit, line 1: for the folder naming, i think views/includes/sic/sic-details.html might be better, then you could group all the sic includes in that folder.

views/includes/task-list/nature-of-business.html
markpit: Matt has done it with the PSC link, have a look at this for an example - https://github.com/companieshouse/confirmation-statement-web/blob/9a10cdfa92e50e9e75b0b71e3bb2336cac93ece3/src/types/task.list.ts#L60

markpit: then you can reference it in the template like this - https://github.com/companieshouse/confirmation-statement-web/blob/main/views/includes/task-list/people.html#L21

src/controllers/confirm.sic.code.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const companyNumber = req.params[urlParams.PARAM_COMPANY_NUMBER];
+    const backLinkUrl = getUrlWithCompanyNumber(TASK_LIST_PATH, companyNumber);
+    return res.render(Templates.SIC, {
+      backLinkUrl
+    });
+  } catch (e) {
+    return next(e);
```
markpit, line 14: to test the exception happening, see if you can do something similar to this test that Jack did - https://github.com/companieshouse/confirmation-statement-web/blob/main/test/controllers/task.list.controller.unit.ts#L87

you could mock the getUrlWithCompanyNumber() function, and tell it to throw an error
`mockGetUrlWithCompanyNumber.mockImplementationOnce(() => {
  throw new Error();
});`

maybe add that mock implementation into your test function so it doesn't happen for the other tests.
Also, no need to check the mockCreateAndLogError in Jack's example as we don't use it here.
Give me a shout if you want a hand.

markpit (approved):

PR #156, opened 2021-06-25
NCS 222 active officers page

LewisJohnEvans:
Create active officers page to be navigated to from task list page


src/controllers/active.officers.controller.ts
```diff
@@ -0,0 +1,10 @@
+import { Request, Response } from "express";
```
kheseyongya, line 1: This file needs to be in the src/controllers/tasks folder

test/controllers/active.officers.controller.unit.ts
```diff
@@ -0,0 +1,38 @@
+import mocks from "../mocks/all.middleware.mock";
```
kheseyongya, line 1: Similar here, this needs to be in the test/controllers/tasks folder.

kheseyongya (approved):

PR #166, opened 2021-07-01
NCS 446 retrieve officer details for active officer page 

LewisJohnEvans:
Implement officer service to retrieve officers details from CH API 

Added feature flag due to lack of officer data in dev environments if true fetch will be made on officer data if false dummy officer data will be used. Config update is ready to be merged if this PR is approved

### DIFF
--- a/src/utils/properties.ts
+++ b/src/utils/properties.ts
@@ -33,4 +33,6 @@ export const INTERNAL_API_URL = getEnvironmentVariable("INTERNAL_API_URL");
 
 export const FEATURE_FLAG_PRIVATE_SDK_12052021 = getEnvironmentVariable("FEATURE_FLAG_PRIVATE_SDK_12052021");
 
+export const FEATURE_FLAG_ACTIVE_OFFICERS_01072021 = getEnvironmentVariable("FEATURE_FLAG_ACTIVE_OFFICERS_01072021", true);
+
 export const PIWIK_START_GOAL_ID = getEnvironmentVariable("PIWIK_START_GOAL_ID");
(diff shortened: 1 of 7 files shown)


LewisJohnEvans:
Added feature flag due to lack of officer data in dev environments if true fetch will be made on officer data if false dummy officer data will be used

src/utils/properties.ts
```diff
@@ -33,4 +33,6 @@ export const INTERNAL_API_URL = getEnvironmentVariable("INTERNAL_API_URL");
 
 export const FEATURE_FLAG_PRIVATE_SDK_12052021 = getEnvironmentVariable("FEATURE_FLAG_PRIVATE_SDK_12052021");
 
+export const FEATURE_FLAG_ACTIVE_OFFICERS_01072021 = getEnvironmentVariable("FEATURE_FLAG_ACTIVE_OFFICERS_01072021", true);
```
ICoxCH, line 36: is this flag neccassary? As we had similar flags for eligibility but that was to allow testers to pass eligibility whilst we were still implementing it.
Here, I don't think we need it, as the service isn't live, and the only people hitting the new page will be the testers checking its working

LewisJohnEvans: Yeah strictly speaking it isn't necessary for this story to pass testing but without it the page isn't reachable in the local journey for devs. So future work displaying the data on the active officers page would need something like this to be implemented.

Im happy to remove it if that's preferred and allow it to be tackled in later stories ?

ICoxCH (approved):

PR #173, opened 2021-07-05
NCS-443 Matomo event tracking on ROA page.

AlfredBwomezi:
Add tracking event on Registered Office Address page.
Also refactored: Extracted "Continue" button to a file rather than have element duplicated across multiple pages.

### DIFF
--- a/views/includes/continue-button.html
+++ b/views/includes/continue-button.html
@@ -1,5 +1,3 @@
-<p class="govuk-body-m">
-
 {{ govukButton({
     text: "Continue",
     attributes: {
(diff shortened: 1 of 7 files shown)


views/includes/continue-button.html
```diff
@@ -0,0 +1,11 @@
+<p class="govuk-body-m">
```
ICoxCH, line 1: is the paragraph tag here neccassary?
As this looks like it will be adding a <p> where there previously wasn't any

ICoxCH (approved):

PR #182, opened 2021-07-06
Ncs 457 display soc data

mattch1:
Web frontend to display statement of capital data.

Unsure what to to do about the transaction id here.


src/services/statement.of.capital.service.ts
```diff
@@ -0,0 +1,15 @@
+import { createOAuthApiClient } from "./api.service";
+import { ConfirmationStatementService, StatementOfCapital
+} from "private-api-sdk-node/dist/services/confirmation-statement";
```
ICoxCH, line 3: this import looks as if its accidentally been formatted wrong

src/controllers/tasks/statement.of.capital.controller.ts
```diff
@@ -3,13 +3,20 @@ import { RADIO_BUTTON_VALUE, STATEMENT_OF_CAPITAL_ERROR } from "../../utils/cons
...
+import { getStatementOfCapitalData } from "../../services/statement.of.capital.service";
+import { Session } from "@companieshouse/node-session-handler";
+import { StatementOfCapital } from "private-api-sdk-node/dist/services/confirmation-statement";
 
-export const get = (req: Request, res: Response, next: NextFunction) => {
+export const get = async(req: Request, res: Response, next: NextFunction) => {
   try {
     const companyNumber = getCompanyNumber(req);
+    const session: Session = req.session as Session;
+    const transactionId: string = "";
```
ICoxCH, line 14: I appreciate that we already need to do a large refactor once the web journey actually creates a transaction.
but presumably the api call following this will fallover without a transaction id. So is there any need to get this pr in prior to the transaction id actually being present?

src/services/statement.of.capital.service.ts
```diff
@@ -0,0 +1,15 @@
...
+import Resource from "@companieshouse/api-sdk-node/dist/services/resource";
+
+export const getStatementOfCapitalData = async (session: Session, transactionId: string, companyNumber: string): Promise<StatementOfCapital> => {
+  const client = createOAuthApiClient(session);
+  const csService: ConfirmationStatementService = client.confirmationStatementService;
+  const response: Resource<StatementOfCapital> = await csService.getStatementOfCapital(transactionId, companyNumber);
+  if (response.httpStatusCode === 200) {
+    return response.resource as StatementOfCapital;
+  }
+  throw new Error("Statement of capital data not found");
```
ICoxCH, line 14: perhaps throw this error only if status == 404.
And then throw a more generic "Error retrieving statement of capital" and include the error from the api call

src/controllers/tasks/statement.of.capital.controller.ts
ICoxCH: or on second thoughts, why does the statement of capital endpoint even need a transaction id?

src/services/statement.of.capital.service.ts
```diff
@@ -0,0 +1,17 @@
+import { createOAuthApiClient } from "./api.service";
+import { ConfirmationStatementService, StatementOfCapital } from "private-api-sdk-node/dist/services/confirmation-statement";
+import { Session } from "@companieshouse/node-session-handler";
+import Resource, { ApiError, ApiErrorResponse } from "@companieshouse/api-sdk-node/dist/services/resource";
+
+export const getStatementOfCapitalData = async (session: Session, companyNumber: string): Promise<StatementOfCapital> => {
+  const client = createOAuthApiClient(session);
+  const csService: ConfirmationStatementService = client.confirmationStatementService;
+  const response: Resource<StatementOfCapital> | ApiErrorResponse = await csService.getStatementOfCapital(companyNumber);
+  if (response.httpStatusCode === 404) {
```
ICoxCH, line 10: this would be better off as >= 400
so it handles all 4XX and 5XX responses as a failed response

src/services/statement.of.capital.service.ts
```diff
@@ -0,0 +1,17 @@
...
+    const errorResponse = response as ApiErrorResponse;
+    const apiErrors = errorResponse.errors as ApiError[];
+    throw new Error("Error retrieving statement of capital " + apiErrors[0].error);
```
ICoxCH, line 13: you could probably just put `JSON.stringify(errorResponse)` put the whole thing in the error to not miss any information

src/services/statement.of.capital.service.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ConfirmationStatementService, StatementOfCapital } from "private-api-sdk-node/dist/services/confirmation-statement";
+import { Session } from "@companieshouse/node-session-handler";
+import Resource, { ApiErrorResponse } from "@companieshouse/api-sdk-node/dist/services/resource";
+
+export const getStatementOfCapitalData = async (session: Session, companyNumber: string): Promise<StatementOfCapital> => {
+  const client = createPrivateOAuthApiClient(session);
+  const csService: ConfirmationStatementService = client.confirmationStatementService;
+  const response: Resource<StatementOfCapital> | ApiErrorResponse = await csService.getStatementOfCapital(companyNumber);
+  const status = response.httpStatusCode as number;
+  if (status >= 404) {
```
ICoxCH, line 11: Should be 400 not 404?
Else its going to treat 400, and 401 etc as a success response and try cast it to a statement of capital

ICoxCH (approved):

PR #184, opened 2021-07-08
NCS-448 Jquery update

mattch1:
Vulnerability update

### DIFF
--- a/views/layout.html
+++ b/views/layout.html
@@ -25,10 +25,11 @@
   <link href="{{assetPath}}/stylesheets/govuk-frontend/v3.11.0/govuk-frontend-ie8-3.11.0.min.css" rel="stylesheet" />
   <![endif]-->
 
-  <script src="https://code.jquery.com/jquery-3.6.0.js"
-          integrity="sha256-WpOohJOqMqqyKL9FccASB9O0KwACQJpFTUBLTYOVvVU="
-          crossorigin="anonymous">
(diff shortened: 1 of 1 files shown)


views/layout.html
```diff
@@ -25,7 +25,7 @@
   <link href="{{assetPath}}/stylesheets/govuk-frontend/v3.11.0/govuk-frontend-ie8-3.11.0.min.css" rel="stylesheet" />
   <![endif]-->
 
-  <script src="https://code.jquery.com/jquery-3.4.1.js"
+  <script src="https://code.jquery.com/jquery-3.6.0.js"
           integrity="sha256-WpOohJOqMqqyKL9FccASB9O0KwACQJpFTUBLTYOVvVU="
```
ICoxCH, line 29: have you checked that the browser actually loads in jquery here?
The integrity hash should have changed between versions. So I expect the browser would block this from running because 3.6.0's sha will be different from this

mattch1: Screen Shot 2021-06-28 at 08 41 27

mattch1: Screen Shot 2021-07-08 at 09 28 57

mattch1: Where do we get the keys from?

mattch1: Updated the hash

ICoxCH (approved):

PR #190, opened 2021-07-16
NCS-449 create submission

SVarallo:
Create a submission after the transaction is created


src/controllers/create.transaction.controller.ts
```diff
@@ -4,16 +4,22 @@ import { urlUtils } from "../utils/url";
...
-    const nextPageUrl = urlUtils.getUrlWithCompanyNumber(TRADING_STATUS_PATH, companyNumber);
-    return res.redirect(nextPageUrl);
+    const transaction: Transaction = await postTransaction(session, companyNumber, DESCRIPTION, REFERENCE);
+    const transactionId = transaction.id as string;
+    const submissionResponse = await createConfirmationStatement(session, transactionId);
+    if (submissionResponse.httpStatusCode === 201) {
+      const nextPageUrl = urlUtils.getUrlWithCompanyNumber(TRADING_STATUS_PATH, companyNumber);
+      return res.redirect(nextPageUrl);
+    }
+    return next(new Error("Company is ineligible " + JSON.stringify(submissionResponse)));
```
markpit, line 22: for the return next, it won't always be the case that it fails due to company being ineligible. Could be a null pointer in the api for example. Also the JSON.stringify can fail if there are circular dependencies in the object. I would just pick out the bits of the object you want to log, status code and the message contents.
try
`    return next(new Error(`Unable to create Confirmation Statement httpStatusCode = ${submissionResponse.httpStatusCode}, resource = ${JSON.stringify(submissionResponse.resource)}`));
`
The resource is an object we've designed so less chance of it having a circular dependency. Would be good to see if it works by changing the api to always throw an error from the create confirmation statement endpoint.

markpit: there are backticks in the code example but github isn't showing them. replace the quotes with backticks
`    return next(new Error("Unable to create Confirmation Statement httpStatusCode = ${submissionResponse.httpStatusCode}, resource = ${JSON.stringify(submissionResponse.resource)}"));
`

src/services/confirmation.statement.service.ts
```diff
@@ -1,11 +1,22 @@
...
 export const createConfirmationStatement = async (session: Session,
-                                                  transactionId: string): Promise<any> => {
+                                                  transactionId: string): Promise<Resource<ConfirmationStatementCreated | CompanyValidationResponse>> => {
   const client = createPrivateOAuthApiClient(session);
   const csService: ConfirmationStatementService = client.confirmationStatementService;
-  return await csService.postNewConfirmationStatement(transactionId);
+  const response = await csService.postNewConfirmationStatement(transactionId);
+  if (response.httpStatusCode !== 201 && response.httpStatusCode !== 400) {
+    const castedResponse: ApiErrorResponse = response;
+    throw new Error("Something went wrong creating confirmation statement " + JSON.stringify(castedResponse));
```
markpit, line 17: might have the same issue here. Digging into it more, the ApiErrorResponse is a CH defined interface so 'hopefully' should not have circular refs, it is usually more the system interfaces that do.
But, I'd at least change the string to use placeholders using the back ticks and ${} 
`    throw new Error("Something went wrong creating confirmation statement ${JSON.stringify(castedResponse)}");
`
replace quotes with backticks

markpit (approved):

PR #197, opened 2021-07-20
NCS-92 Fetch active officer details data from cs-api.

AlfredBwomezi:


src/services/active.officer.details.service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const getActiveOfficerDetailsData = async (session: Session, companyNumber: string): Promise<ActiveOfficerDetails> => {
+  const client = createPrivateOAuthApiClient(session);
+  const csService: ConfirmationStatementService = client.confirmationStatementService;
+  const response: Resource<ActiveOfficerDetails> | ApiErrorResponse = await csService.getActiveOfficerDetails(companyNumber);
+  const status = response.httpStatusCode as number;
+
+  if (status >= 400) {
+    const errorResponse = response as ApiErrorResponse;
+    throw new Error("Error retrieving active officer details: " + JSON.stringify(errorResponse));
```
markpit, line 15: you can use placeholders to build the string here (using the backticks and ${}), i think it reads better - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

markpit (approved):

PR #203, opened 2021-07-20
Ncs 506 update service

mattch1:
In order to get the statement of capital data into the post function I have had to pass in a json string from get and place it into a hidden input field so that it appears in the req body of the post function.


mattch1:
Closing this as we are going to use a cookie instead to pass the soc data

PR #204, opened 2021-07-21
Ncs 460 display active officer details

AlfredBwomezi:


test/controllers/tasks/active.officers.controller.unit.ts
```diff
@@ -1,77 +1,100 @@
...
 
-  it("Should skip officer check and return active officers page if feature flag is off", async () => {
-    mockIsActiveFeature.mockReturnValueOnce(false);
-    const response = await request(app).get(url);
-    expect(response.text).toContain(EXPECTED_TEXT);
-  });
+      expect(response.text).toContain(PAGE_HEADING);
+      expect(response.text).toContain("Is the director still active?");
+      expect(response.text).toContain(mockActiveOfficerDetails.foreName1);
+      expect(response.text).toContain(mockActiveOfficerDetails.dateOfBirth);
```
markpit, line 46: could you add an expect to check forename2 appears in the response.text and add another test to run with no forename2 (undefined). For the second test you could just set mockActiveOfficerDetails.forname2 to undefined and set it back to it's original value at the end of that test. (you'll have to store the value in a const so you can set forename2 to that value at end of test).

test/controllers/tasks/active.officers.controller.unit.ts
```diff
@@ -1,77 +1,117 @@
...
+
+      mockGetActiveOfficerDetails.mockResolvedValueOnce(mockActiveOfficerDetails);
+      const response = await request(app).get(ACTIVE_OFFICER_DETAILS_URL);
+
+      expect(response.text).toContain(PAGE_HEADING);
+      expect(response.text).toContain("Is the director still active?");
+      expect(response.text).toContain(mockActiveOfficerDetails.foreName1);
+      expect(response.text).toContain(mockActiveOfficerDetails.dateOfBirth);
+      expect(response.text).toContain(mockActiveOfficerDetails.nationality);
+
```
markpit, line 63: i think you just need one more expect(response.text).**not**.toContain(mockActiveOfficerDetails.foreName2);
just to check it is not on the page. That should be it then!

markpit (approved):

markpit (approved):

PR #205, opened 2021-07-21
NCS-480 Adding get confirmation statement function to confirmation st…

markpit:
…atement service

Adding get confirmation statement function to confirmation statement service to return a confirmation statement for the given id (and transaction id).
This calls the confirmation-statement-api via the private-api-sdk-node to return an object representing the Mongo record.


test/mocks/statement.of.capital.mock.ts
```diff
@@ -3,11 +3,10 @@ import { StatementOfCapital } from "private-api-sdk-node/dist/services/confirmat
 export const mockStatementOfCapital: StatementOfCapital = {
   classOfShares: "Ordinary",
   currency: "GBP",
-  numberAllotted: 10,
-  aggregateNominalValue: 1,
+  numberAllotted: "10",
```
markpit, line 6: spotted these were the wrong types, numbers instead of strings and totalCurrency doesn't exist on the StatementOfCapital so removed it.

test/services/confirmation.statement.service.unit.ts
```diff
@@ -2,58 +2,118 @@ jest.mock("private-api-sdk-node/dist/services/confirmation-statement");
...
 describe ("Confirmation statement api service unit tests", () => {
 
   beforeEach (() => {
     jest.clearAllMocks();
   });
 
-  it ("should call create confirmation statement in the private sdk", async () => {
-    mockPostNewConfirmationStatement.mockResolvedValueOnce({
-      httpStatusCode: 201
+  describe ("createConfirmationStatement unit tests", () => {
```
markpit, line 36: basically added sub-describes to group tests per function

ICoxCH (approved):

PR #209, opened 2021-07-22
NCS-506 Update cs with soc

mattch1:
Update confirmation statement.


src/controllers/tasks/statement.of.capital.controller.ts
```diff
@@ -5,8 +5,12 @@ import { Templates } from "../../types/template.paths";
 import { urlUtils } from "../../utils/url";
 import { getStatementOfCapitalData } from "../../services/statement.of.capital.service";
 import { Session } from "@companieshouse/node-session-handler";
-import { StatementOfCapital } from "private-api-sdk-node/dist/services/confirmation-statement";
+import {
+  ConfirmationStatementSubmission, SectionStatus,
```
SVarallo, line 9: Should SectionStatus be on its own line?

src/controllers/tasks/statement.of.capital.controller.ts
```diff
@@ -55,4 +60,27 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
   }
 };
 
+const sendUpdate = async (transactionId: string, submissionId: string, req: Request) => {
+  const statementOfCapital: StatementOfCapital = req.sessionCookie["statementOfCapital"];
```
markpit, line 64: could you add a 'sessionCookieConstants' exported object in constants.ts and add the "statementOfCapital" to it and use it on line 64 then. That way, we have a central place to store the session cookie key names.

src/services/confirmation.statement.service.ts
```diff
@@ -37,3 +37,18 @@ export const getConfirmationStatement = async (session: Session, transactionId:
 
   return csSubmissionResource.resource;
 };
+
+export const updateConfirmationStatement = async (session: Session,
```
markpit, line 41: it doesn't look like the returned value from updateConfirmationStatement gets used, I don't think we'd need to return anything from the update so you can remove the return response and the return type.

test/mocks/statement.of.capital.mock.ts
```diff
@@ -1,12 +1,28 @@
...
-  totalAmountUnpaidForCurrency: "1"
+  numberAllotted: "100",
+  aggregateNominalValue: "0.01",
+  prescribedParticulars: "THE QUICK BROWN FOX",
+  totalNumberOfShares: "100",
+  totalAggregateNominalValue: "1",
+  totalAmountUnpaidForCurrency: "2"
+};
+
+export const mockConfirmationStatementSubmission: ConfirmationStatementSubmission = {
```
markpit, line 17: there is a mockConfirmationStatementSubmission in confirmation.statement.submission.mock. Looks very similar, could you move this mock into the confirmation.statement.submission.mock.ts file with the other and possibly re-use / combine with the existing one?

markpit (approved):

PR #219, opened 2021-07-26
Configure cookie session

ICoxCH:


mattch1 (approved):

src/app.ts
```diff
@@ -38,12 +39,9 @@ app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
 
 // apply middleware
-// TODO get key and timeout from config
 app.use(cookieSessionMiddleware({
-  keys: ["TO BE CHANGED"],
-  // Cookie Options
-  maxAge: 60 * 60 * 1000 // 1 hours
-} as any) as any);
+  keys: [COOKIE_SECRET]
```
markpit, line 43: should we leave the maxAge in until we add it to config?

ICoxCH: Max age isnt used by the middleware as far as i can see

markpit (approved):

PR #222, opened 2021-07-27
NCS-535 Fix: Active officer details page broken back button

AlfredBwomezi:
Fix the broken navigation links on the page.
Fix the page details that were disappearing on errorMessage page render.

### DIFF
--- a/src/controllers/tasks/active.officers.controller.ts
+++ b/src/controllers/tasks/active.officers.controller.ts
@@ -11,6 +11,8 @@ import { formatTitleCase } from "../../utils/format";
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
     const companyNumber = getCompanyNumber(req);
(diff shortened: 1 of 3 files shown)


src/controllers/tasks/active.officers.controller.ts
```diff
@@ -8,19 +8,26 @@ import { ActiveOfficerDetails } from "private-api-sdk-node/dist/services/confirm
 import { getActiveOfficerDetailsData } from "../../services/active.officer.details.service";
 import { formatTitleCase } from "../../utils/format";
 
+let activeOfficerDetails: ActiveOfficerDetails;
```
markpit, line 11: don't think we can use global variables, get and post will be coming in on different requests, also with scaling/load balancing, get and post could happen on different servers

markpit (approved):

PR #226, opened 2021-07-28
NCS-552 Template for RLE

mattch1:
Until we have the data back from the api, have created a template with hardcoded values and hardcoded a flag in the controller so that the current functionality is not changed. Created a subtask for this: https://companieshouse.atlassian.net/browse/NCS-553. Will unit teest this as well on this next subtask as the test will require a mock model object.

Also need to fix CDN rendering so will subtask this as well.


views/tasks/includes/relevant-legal-entity.html
```diff
@@ -0,0 +1,64 @@
+{% set pscHTML %}
+<div class="govuk-summary-card govuk-!-margin-bottom-4">
+    <div class="govuk-summary-card_heading govuk-!-padding-top-4 govuk-!-padding-bottom-4 govuk-!-padding-right-3 govuk-!-padding-left-3" style="display: flex; justify-content: space-between; align-items: center">
+        <p class="govuk-!-font-weight-bold govuk-!-margin-0">BROOKLYN NETS LIMITED</p>
+        {{govukTag({
```
markpit, line 5: bit of indenting needed

src/controllers/tasks/people.with.significant.control.controller.ts
```diff
@@ -23,13 +24,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
   }
 };
 
+const getTemplate = (isRle): string => {
+  if (isRle) {
+    return Templates.PEOPLE_WITH_SIGNIFICANT_CONTROL_RLE;
```
markpit, line 29: as we can't unit test this new code and it can't be reached yet due to the 'false' I'd say don't commit the controller code,  just have the template in the PR.

markpit (approved):

PR #240, opened 2021-08-02
NCS-360 save sic code data to database

kheseyongya:
[NCS-360](https://companieshouse.atlassian.net/browse/NCS-360)

Update the tag in task list page according to the data from the database.


markpit:
I've agreed with Alfred that the duplication can be tackled in a separate PR as it will affect other controllers as well as this one.

markpit (approved):

PR #246, opened 2021-08-04
NCS 377 save ROA data

LewisJohnEvans:
submit ROA data to post CS update after review ROA details page


src/controllers/tasks/registered.office.address.controller.ts
```diff
@@ -23,22 +30,27 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
   }
 };
 
-export const post = (req: Request, res: Response, next: NextFunction) => {
+export const post = async (req: Request, res: Response, next: NextFunction) => {
   try {
     const roaButtonValue = req.body.registeredOfficeAddress;
+    const transactionId = req.params[urlParams.PARAM_TRANSACTION_ID];
```
markpit, line 36: have a chat to alfred, you should be able to replace the req.params calls with those in the utils/url.ts file.
import urlUtils i think.

markpit: alfred moved some of them into there

markpit (approved):

PR #259, opened 2021-08-09
NCS-581 Validation error psc statment

mattch1:


src/controllers/tasks/psc.statement.controller.ts
```diff
@@ -18,3 +19,17 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     return next(e);
   }
 };
+
+export const post = (req: Request, res: Response) => {
+  const pscButtonValue = req.body.pscStatementValue;
```
markpit, line 24: could you wrap this in a try catch and return next(e) in the catch. This wil enable us to show the error page if any exceptions get thrown.

markpit: i mean the contents of the post function

test/controllers/tasks/psc.statement.controller.unit.ts
```diff
@@ -42,4 +44,15 @@ describe("PSC Statement controller tests", () => {
       spyGetUrlToPath.mockRestore();
     });
   });
+
+  describe("post tests", function () {
+    it("Should redisplay psc page with error when radio button is not selected", async () => {
+      const response = await request(app).post(PSC_STATEMENT_URL);
+
+      expect(response.status).toEqual(200);
```
markpit, line 52: can you add a test to check the error gets thrown similar to what was done in the get function

markpit (approved):

PR #274, opened 2021-08-11
Ncs 607 handle blank dob

mattch1:
If rle return blank and proceed to page.
If psc and date of birth return date of birth and proceed to page.
If no psc and date of birth throw error with message containing surname of psc.


src/controllers/tasks/people.with.significant.control.controller.ts
```diff
@@ -118,8 +118,18 @@ const updateCsSubmission = (currentCsSubmission: ConfirmationStatementSubmission
...
 };
+
+const handleDateOfBirth = (pscAppointmentType: string, psc: PersonOfSignificantControl): string => {
+  if (pscAppointmentType === appointmentTypeNames.RLE) {
+    return "";
+  }
+  if (psc.dateOfBirth) {
+    return toReadableFormatMonthYear(psc.dateOfBirth.month, psc.dateOfBirth.year);
+  }
+  throw createAndLogError(`Date of birth missing for individual psc ${psc.nameElements.surname}`);
```
markpit, line 134: can you add an undefined check to ${psc.nameElements.surname} like so ${psc.nameElements?.surname}
this will return undefined if nameElements is undefined. Otherwise it will throw a null pointer when trying to access surname on undefined and will fail to log the message. With the'?' in place, the message should display but with 'undefined' in the message placeholder.

markpit (approved):

PR #282, opened 2021-08-13
Refactor sendUpdate and updateCsSubmission duplicated methods

CHJSkinner:
Refactor send update and updateCsSubmission duplicated methods


src/utils/update.confirmation.statement.submission.ts
```diff
@@ -0,0 +1,64 @@
...
+      const newSicData: SicCodeData = {
+        sectionStatus: status
+      };
+      currentCsSubmission.data.sicCodeData = newSicData;
+      break;
+    case SECTIONS.SOC:
+      const newSocData: StatementOfCapitalData = {
+        sectionStatus: status,
+      };    
+      currentCsSubmission.data.statementOfCapitalData = newSocData;
```
markpit, line 59: for SOC we also add the 'statementOfCapital' section as well as the sectionStatus
from statement.of.capital.controller.ts - 
const newSocData: StatementOfCapitalData = {
    sectionStatus: status,
    ...(statementOfCapital && { statementOfCapital: statementOfCapital })
  };

i think it would be better if the controllers passed in the data to be set eg this function would be:
const updateCsSubmission = (currentCsSubmission: ConfirmationStatementSubmission, sectionName: SECTIONS,  sectionData: any): ConfirmationStatementSubmission => {

if (!currentCsSubmission.data) {
    currentCsSubmission.data = {};
  }

currentCsSubmission.data[sectionName] = sectionData;

}

SECTIONS would have to be modified to contain the fieldNames of the sections ie same as in the ConfirmationStatementSubmission interface =
    personsSignificantControlData
    statementOfCapitalData
    sicCodeData
    registeredOfficeAddressData
    activeDirectorDetailsData


Then teh controller can construct the data it wants to add and pass it into this updateCsSubmission function like so
const newSocData: StatementOfCapitalData = {
    sectionStatus: status,
    ...(statementOfCapital && { statementOfCapital: statementOfCapital })
  };
updateCsSubmission(currentCsSubmission, SECTIONS.SOC, newSocData);

CHJSkinner: Followed suggestions and updated methods as such - Resolved

markpit (approved):

PR #295, opened 2021-08-25
NCS-641 add PUT functionality to transaction service

markpit:
Updating the transaction.service to use the new PUT transaction functionality in the api-sdk-node.
Also added a function to handle closing a transaction.
Updated the transaction unit tests and fixed a few of the error checking ones which were passing with the expect failing.

### DIFF
--- a/src/services/transaction.service.ts
+++ b/src/services/transaction.service.ts
@@ -4,7 +4,8 @@ import { createAndLogError, logger } from "../utils/logger";
 import { createPublicOAuthApiClient } from "./api.service";
 import { Session } from "@companieshouse/node-session-handler";
 import ApiClient from "@companieshouse/api-sdk-node/dist/client";
-import { ApiErrorResponse } from "@companieshouse/api-sdk-node/dist/services/resource";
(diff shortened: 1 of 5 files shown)


src/services/transaction.service.ts
```diff
@@ -30,10 +31,62 @@ export const postTransaction = async (session: Session, companyNumber: string, d
...
   logger.debug(`Received transaction ${JSON.stringify(sdkResponse)}`);
 
   return castedSdkResponse.resource;
 };
+
+/**
+ * Response can contain a URL to start payment session if payment is needed
+ */
+export const closeTransaction = async (session: Session, companyNumber: string, transactionId: string): Promise<string | undefined> => {
+  const apiResponse: ApiResponse<Transaction> = await putTransaction(session, companyNumber, transactionId, DESCRIPTION, REFERENCE, transactionStatuses.CLOSED);
```
markpit, line 46: I've gone with the default transaction description and reference here:
DESCRIPTION = "Confirmation Statement Transaction";
REFERENCE = "ConfirmationStatementReference";
not sure if this should be something more specific?

PR #302, opened 2021-08-31
NCS-82 Add event log for registers page

kheseyongya:
[NCS-82](https://companieshouse.atlassian.net/secure/RapidBoard.jspa?rapidView=234&projectKey=BI&modal=detail&selectedIssue=NCS-82)

### DIFF
--- a/views/tasks/register-locations.html
+++ b/views/tasks/register-locations.html
@@ -65,7 +65,7 @@
                 }
               },
               {
-                value: "yes",
+                value: "recently_filed",
                 text: "No, but i've recently filed an update",
                 attributes: {
                   "data-event-id": "recently-filed-radio-button"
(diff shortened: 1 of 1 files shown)


views/tasks/register-locations.html
```diff
@@ -56,15 +56,24 @@ <h1 class="govuk-heading-l">Check where the company records are kept</h1>
...
               {
                 value: "no",
-                text: "No"
+                text: "No",
+                attributes: {
+                  "data-event-id": "no-radio-button"
+                }
               },
               {
                 value: "yes",
```
markpit, line 72: value should be "recently_filed" for recently filed buttons

markpit:
the duplication looks like it's caused by {% if registers === 'yes' %}, no etc..  which is replicating the radio buttons to make them checked. I can't see 'registers' being passed in from the controller so you can probably remove the duplicated blocks and just have one.

markpit (approved):

PR #317, opened 2021-09-03
NCS 338 display register location data

LewisJohnEvans:
fetch and display register location data on the register location page

### DIFF
--- a/src/controllers/tasks/register.locations.controller.ts
+++ b/src/controllers/tasks/register.locations.controller.ts
@@ -12,10 +12,8 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
     const companyNumber = urlUtils.getCompanyNumberFromRequestParams(req);
     const session: Session = req.session as Session;
     const registerLocations: RegisterLocation[] = await getRegisterLocationData(session, companyNumber);
-    console.log("JSON OBJECT: " + JSON.stringify(registerLocations));
     const sailAddress = formatSailForDisplay(registerLocations);
     const registers = formatRegisterForDisplay(registerLocations);
-    console.log("registers: " + registers + " length: " + registers.length);
     const backLinkUrl = urlUtils.getUrlToPath(TASK_LIST_PATH, req);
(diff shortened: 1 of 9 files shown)


src/controllers/tasks/register.locations.controller.ts
```diff
@@ -40,3 +49,21 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
...
+  let sailAddress = "";
+  if (regLoc.length && regLoc[0].sailAddress) {
+    sailAddress = formatAddressForDisplay(formatAddress(regLoc[0].sailAddress));
+  }
+  return sailAddress;
+};
+
+const formatRegisterForDisplay = (regLoc: RegisterLocation[]): Array<string> => {
+  const registers = new Array<string>();
+  for (const key of Object.values(regLoc)) {
```
markpit, line 63: i'd rename the key var to value as you are iterating through the object.values here

src/services/register.location.service.ts
```diff
@@ -0,0 +1,18 @@
...
+import { createPrivateOAuthApiClient } from "./api.service";
+
+export const getRegisterLocationData = async (session: Session, companyNumber: string): Promise<RegisterLocation[]> => {
+  const client = createPrivateOAuthApiClient(session);
+  const csService: ConfirmationStatementService = client.confirmationStatementService;
+  const response: Resource<RegisterLocation[]> | ApiErrorResponse = await csService.getRegisterLocations(companyNumber);
+  const status = response.httpStatusCode as number;
+  if (status >= 400) {
+    const errorResponse = response as ApiErrorResponse;
+    throw new Error(`Error retrieving register location data from confirmation-statment api: ${JSON.stringify(errorResponse)}`);
```
markpit, line 14: use 
throw createAndLogError(`Error retrieving register location data from confirmation-statment api: ${JSON.stringify(errorResponse)}`);
instead, it will log the error as well.
from "../../utils/logger"

markpit (approved):

PR #333, opened 2021-09-09
NCS-646 Add validation check for statement-of-capital totalAmountUnpaid.

AlfredBwomezi:
Validate statement-of-capital totalAmountUnpaidForCurrency by checking that it is not NULL.

### DIFF
--- a/src/controllers/tasks/statement.of.capital.controller.ts
+++ b/src/controllers/tasks/statement.of.capital.controller.ts
@@ -22,7 +22,7 @@ export const get = async(req: Request, res: Response, next: NextFunction) => {
     const session: Session = req.session as Session;
     const statementOfCapital: StatementOfCapital = await getStatementOfCapitalData(session, companyNumber);
     const sharesValidation = await validateTotalNumberOfShares(session, companyNumber, +statementOfCapital.totalNumberOfShares);
-    const totalAmountUnpaidValidation = statementOfCapital.totalAmountUnpaidForCurrency !== null;
+    const totalAmountUnpaidValidation = typeof statementOfCapital.totalAmountUnpaidForCurrency === 'string';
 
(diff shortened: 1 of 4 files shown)


src/controllers/tasks/statement.of.capital.controller.ts
```diff
@@ -22,6 +22,7 @@ export const get = async(req: Request, res: Response, next: NextFunction) => {
     const session: Session = req.session as Session;
     const statementOfCapital: StatementOfCapital = await getStatementOfCapitalData(session, companyNumber);
     const sharesValidation = await validateTotalNumberOfShares(session, companyNumber, +statementOfCapital.totalNumberOfShares);
+    const totalAmountUnpaidValidation = statementOfCapital.totalAmountUnpaidForCurrency !== null;
```
markpit, line 25: statementOfCapital.totalAmountUnpaidForCurrency can also be undefined (more likely than null). To check if it is present try using       typeof statementOfCapital.totalAmountUnpaidForCurrency === 'string' instead

test/controllers/tasks/statement.of.capital.controller.unit.ts
```diff
@@ -56,7 +57,40 @@ describe("Statement of Capital controller tests", () => {
   });
 
   describe("get tests", () => {
-    it("should navigate to the statement of capital page with NO buttons visible if share totals DO NOT match the shareholders'.", async () => {
+
+    it("should navigate to the statement of capital page with NO buttons visible, a share-capital-mismatch-warning and total-amount-unpaid-warning messages if Share totals DO NOT match and amount unpaid is NULL", async () => {
+      mockGetStatementOfCapitalData.mockResolvedValueOnce({
```
markpit, line 62: I'd also add a test to do the same as this one but with amount unpaid set to undefined, so that we know the code will handle it being undefined.

markpit (approved):

PR #336, opened 2021-09-09
NCS-112 create Confirmation page

SVarallo:


test/controllers/confirmation.controller.unit.ts
```diff
@@ -0,0 +1,46 @@
+jest.mock("../../src/services/confirmation.statement.service");
```
markpit, line 1: i think this can be removed, can't see it being used in the controller

markpit (approved):

PR #340, opened 2021-09-10
correct return uri for payments journey

ICoxCH:


src/services/payment.service.ts
```diff
@@ -6,13 +6,19 @@ import { v4 as uuidv4 } from "uuid";
...
 import { API_URL, CHS_URL } from "../utils/properties";
+import { CONFIRMATION_PATH, urlParams } from "../types/page.urls";
 
 export const startPaymentsSession = async (session: Session, paymentSessionUrl: string,
-                                           paymentResourceUri: string, filingResourceUri: string): Promise<ApiResponse<Payment>> => {
+                                           paymentResourceUri: string, submissionId: string, transactionId: string, companyNumber: string): Promise<ApiResponse<Payment>> => {
   const apiClient: ApiClient = createPaymentApiClient(session, paymentSessionUrl);
   const resourceWithHost = API_URL + paymentResourceUri;
+  const redirectUri: string = `${CHS_URL}${CONFIRMATION_PATH}`
+    .replace(`:${urlParams.PARAM_TRANSACTION_ID}`, transactionId)
```
markpit, line 16: can you use the (aptly named!) urlUtils function getUrlWithCompanyNumberTransactionIdAndSubmissionId
here instead of the replace, it'll do the same thing

markpit (approved):

PR #362, opened 2021-09-16
NCS-676 Payment screen navigation

mattch1:
Redirect from submitted landing screen.


src/controllers/confirmation.controller.ts
```diff
@@ -1,13 +1,25 @@
...
-    });
+    const paymentStatus = req.query.status;
+    if (paymentStatus === "paid") {
+      const transactionId = urlUtils.getTransactionIdFromRequestParams(req);
+      return res.render(Templates.CONFIRMATION, {
+        referenceNumber: transactionId,
+      });
+    } else {
+      logger.info("Payment status: " + paymentStatus + " - redirecting to the review page");
+      const companyNumber = urlUtils.getCompanyNumberFromRequestParams(req);
```
markpit, line 17: there is a shortcut way of building the url by using urlUtils.getUrlToPath and pass in the req object. Should reduce the number of lines then.

ICoxCH:
would it make more sense for us to pass in the url for a new controller just to handle the payment callback?
And redirect to the appropriate page from there?
Something like /conf-statement/{id}/transaction/{id}/submission/{id}/payment-callback
So we don't clutter the confirmation controller with handling a payment redirect?

markpit:
> would it make more sense for us to pass in the url for a new controller just to handle the payment callback?
> And redirect to the appropriate page from there?
> Something like /conf-statement/{id}/transaction/{id}/submission/{id}/payment-callback
> So we don't clutter the confirmation controller with handling a payment redirect?

so the payment service "return to" url would be a new payment-callback url/controller in our web app and that would do the 'paid' check and redirect the user to the appropriate page? it would be good to do that, makes it more flexible if they want to redirect to a different page in future.

src/controllers/payment.callback.controller.ts
```diff
@@ -0,0 +1,19 @@
+import { NextFunction, Request, Response } from "express";
+import { urlUtils } from "../utils/url";
+import { logger } from "../utils/logger";
+import { CONFIRMATION_PATH, REVIEW_PATH } from "../types/page.urls";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const paymentStatus = req.query.status;
+    if (paymentStatus === "paid") {
+      logger.info("Payment status: " + paymentStatus + " - redirecting to the confirmation page");
```
ICoxCH, line 10: Could we either add some sort of identifier to these logs (submission id maybe?) or set them to debug

ICoxCH (approved):

PR #366, opened 2021-09-20
NCS-684 Remove session cookies

kheseyongya:
[NCS-684](https://companieshouse.atlassian.net/secure/RapidBoard.jspa?rapidView=234&projectKey=NCS&modal=detail&selectedIssue=NCS-684)

Remove the use of session cookies for retrieving the task data and recall the api instead on the post function.


src/controllers/tasks/registered.office.address.controller.ts
```diff
@@ -30,8 +29,9 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
 
 export const post = async (req: Request, res: Response, next: NextFunction) => {
   try {
+    const companyNumber = urlUtils.getCompanyNumberFromRequestParams(req);
```
markpit, line 32: you can move lines 32,33 to just before 50, it will save us an api call radio button is yes or no

kheseyongya: amended it now

markpit (approved):

PR #370, opened 2021-09-22
NCS-268 Use made up to date from CS Submission on TaskList page.

AlfredBwomezi:


src/services/task.list.service.ts
```diff
@@ -57,7 +55,7 @@ export const initTaskList = (companyNumber: string,
 
   return {
     tasks: allTasks,
-    recordDate: toReadableFormat(DateTime.now().toString()),
+    recordDate: csSubmission.data?.confirmationStatementMadeUpToDate,
```
markpit, line 58: i think we'll still need to convert it to readable format here, it should appear on screen like 15 September 2021 rather than 2021-09-15

markpit (approved):

PR #373, opened 2021-09-22
Ncs 690 check returned payment state

ICoxCH:


src/controllers/payment.callback.controller.ts
```diff
@@ -1,12 +1,23 @@
...
 import { CONFIRMATION_PATH, REVIEW_PATH } from "../types/page.urls";
+import { Session } from "@companieshouse/node-session-handler";
 
 export const get = (req: Request, res: Response, next: NextFunction) => {
   try {
+    const session = req.session as Session;
     const paymentStatus = req.query.status;
     const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
+    const returnedState = req.query.state;
+    const state = session.getExtraData("payment-nonce");
```
markpit, line 13: could you rename the state const to savedState? think it might be clearer that way

markpit (approved):

PR #380, opened 2021-09-23
Fix address arrangement to march specified order in NCS-714

AlfredBwomezi:
The correct order of the address data to be displayed on web is;

> care_of
> po_box
> premises
> address_line_1
> address_line_2
> locality
> region
> country
> postal_code


src/controllers/tasks/registered.office.address.controller.ts
```diff
@@ -63,13 +63,13 @@ const formatAddress = (address: RegisteredOfficeAddress): RegisteredOfficeAddres
   const addressClone: RegisteredOfficeAddress = JSON.parse(JSON.stringify(address));
   return {
     careOf: formatTitleCase(addressClone.careOf),
+    poBox: formatTitleCase(addressClone.poBox),
```
markpit, line 66: can you move the formatAddress function into the fomat.ts module? may have to rename the function if it clashes with existing formatAddress

markpit (approved):

PR #421, opened 2021-11-08
Ncs 723 natural person secretaries details page

CHJSkinner:


views/tasks/natural-person-secretaries.html
```diff
@@ -0,0 +1,93 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+{% if errorList.length > 0 %}
+  Error:
+{% endif %}  Review the officers
```
markpit, line 6: just wondering if it should be Review the secretaries as page title?

CHJSkinner: Spoke to analyst and agreed with change of page title, updated

markpit (approved):

PR #453, opened 2021-11-24
BI-9488 Initial change for navigating through the officers page

kheseyongya:
[BI-9488](https://companieshouse.atlassian.net/jira/software/c/projects/NCS/boards/234?modal=detail&selectedIssue=BI-9488)

Navigate to the relevant officers page depending on the roles present in the officers list.

This is for navigating to and from natural persons secretary page.


src/controllers/tasks/natural.person.secretaries.controller.ts
```diff
@@ -31,9 +31,31 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
     const transactionId = urlUtils.getTransactionIdFromRequestParams(req);
     const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
     const session: Session = req.session as Session;
+    const officers: ActiveOfficerDetails[] = await getActiveOfficersDetailsData(session, transactionId, submissionId);
     const natPersonSecretariesBtnValue = req.body.naturalPersonSecretaries;
+
     if (natPersonSecretariesBtnValue === RADIO_BUTTON_VALUE.YES || natPersonSecretariesBtnValue === RADIO_BUTTON_VALUE.RECENTLY_FILED) {
-      // TODO update with correct next page in journey (not task list)
+      let corpSecretary = false;
```
markpit, line 38: for the sonar code smell for this method you can split the if-else if -else up
the elseif part i would turn into a 'guard clause' which you can do early in this function
you could split them into if-then-return blocks like this without using 'else'

`if (natPersonSecretariesBtnValue === RADIO_BUTTON_VALUE.NO) {
      return res.render(Templates.WRONG_DETAILS, {
        templateName: Templates.WRONG_DETAILS,
        backLinkUrl: urlUtils.getUrlToPath(NATURAL_PERSON_SECRETARIES_PATH, req),
        returnToTaskListUrl: urlUtils.getUrlToPath(TASK_LIST_PATH, req),
        stepOneHeading: WRONG_DETAILS_UPDATE_SECRETARY,
        pageHeading: WRONG_DETAILS_UPDATE_OFFICERS,
      });`

`if (natPersonSecretariesBtnValue === RADIO_BUTTON_VALUE.YES || natPersonSecretariesBtnValue === RADIO_BUTTON_VALUE.RECENTLY_FILED) {
      let corpSecretary = false;
      let director = false;
      let corpDirector = false;
      officers.forEach(officer => {
        if (officer.role === "SECRETARY" && officer.isCorporate){
          corpSecretary = true;
        } else if (officer.role === "DIRECTOR" && !officer.isCorporate){
          director = true;
        } else if (officer.role === "DIRECTOR" && officer.isCorporate){
          corpDirector = true;
        }
      });

      if (corpSecretary){
        return res.redirect(urlUtils.getUrlToPath(CORPORATE_SECRETARIES_PATH, req));
      } else if (director){
        return res.redirect(urlUtils.getUrlToPath(NATURAL_PERSON_DIRECTORS_PATH, req));
      } else if (corpDirector){
        return res.redirect(urlUtils.getUrlToPath(CORPORATE_DIRECTORS_PATH, req));
      }

      return res.redirect(urlUtils.getUrlToPath(TASK_LIST_PATH, req));
    }`

then the final part wouldn't need to be wrapped in an else as the other 'ifs' above would return if they were entered

`const officers: ActiveOfficerDetails[] = await getActiveOfficersDetailsData(session, transactionId, submissionId);
      const secretaryList = formatSecretaryList(officers);
      return res.render(Templates.NATURAL_PERSON_SECRETARIES, {
        secretaryErrorMsg: SECRETARY_DETAILS_ERROR,
        templateName: Templates.NATURAL_PERSON_SECRETARIES,
        secretaryList
      });`

markpit: formatting hasn't come out very well in my comment sorry

markpit: hopefully that should reduce the sonar complexity rating

src/controllers/tasks/active.officers.details.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+import { CORPORATE_DIRECTORS_PATH, CORPORATE_SECRETARIES_PATH, NATURAL_PERSON_DIRECTORS_PATH, NATURAL_PERSON_SECRETARIES_PATH } from "../../types/page.urls";
+import { urlUtils } from "../../utils/url";
+
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const transactionId = urlUtils.getTransactionIdFromRequestParams(req);
+    const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
+    const session: Session = req.session as Session;
+    const officers: ActiveOfficerDetails[] = await getActiveOfficersDetailsData(session, transactionId, submissionId);
+    let secretary = false;
```
markpit, line 14: i'd rename these vars to containsSecretary, containsDirector etc...
I found it a bit confusing when i first read it as i thought we were checking what type the each officer was, but realised we are saying 'if the array contains a secretary then go to the secretary screen first'

src/controllers/tasks/active.officers.details.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+    let corpSecretary = false;
+    let director = false;
+    let corpDirector = false;
+
+    officers.forEach(officer => {
+      if (officer.role === "SECRETARY" && !officer.isCorporate){
```
markpit, line 20: also, the "SECRETARY" and "DIRECTOR" are excellent candidates to be defined in the constants.ts module , possibly as enums?

src/services/task.list.service.ts
```diff
@@ -26,7 +24,7 @@ export const initTaskList = (companyNumber: string,
   const allTasks = {
     officers: {
       state: toTaskState(csSubmission.data?.activeOfficerDetailsData?.sectionStatus),
-      url: urlUtils.getUrlWithCompanyNumberTransactionIdAndSubmissionId(officerSection(), companyNumber, transactionId, submissionId)
+      url: urlUtils.getUrlWithCompanyNumberTransactionIdAndSubmissionId(ACTIVE_OFFICERS_DETAILS_PATH, companyNumber, transactionId, submissionId)
```
markpit, line 27: i think we still need the feature flag here as this looks like it will always go to the 'new' officers details path. 
maybe have the officerSection function back in but have it return ACTIVE_OFFICERS_DETAILS_PATH instead of NATURAL_PERSON_SECRETARIES_PATH like
`const officerSection = (): string => {
  if (FEATURE_FLAG_FIVE_OR_LESS_OFFICERS_JOURNEY_21102021 === 'true') {
    return ACTIVE_OFFICERS_DETAILS_PATH;
  } else {
    return ACTIVE_OFFICERS_PATH;
  }
};`

markpit (approved):

PR #463, opened 2021-11-29
Change constants to lower case to match the retrieved fields

kheseyongya:
[NCS-9488
](https://companieshouse.atlassian.net/jira/software/c/projects/NCS/boards/234?modal=detail&selectedIssue=BI-9488)


src/utils/format.ts
```diff
@@ -94,7 +94,7 @@ export const formatPSCForDisplay = (psc: PersonOfSignificantControl): PersonOfSi
 export const formatSecretaryList = (officers: ActiveOfficerDetails[]): any[] => {
   const secretaryList = new Array(0);
   for (const officer of officers) {
-    if (officer.role === "SECRETARY" && officer.isCorporate === false ) {
+    if (officer.role === "Secretary" && officer.isCorporate === false ) {
```
ICoxCH, line 97: why not use the enum OFFICER_ROLE? `officer.role === OFFICER_ROLE.SECRETARY`?

kheseyongya: ah completely forgot about it. will amend it

src/utils/format.ts
```diff
@@ -94,7 +95,7 @@ export const formatPSCForDisplay = (psc: PersonOfSignificantControl): PersonOfSi
 export const formatSecretaryList = (officers: ActiveOfficerDetails[]): any[] => {
   const secretaryList = new Array(0);
   for (const officer of officers) {
-    if (officer.role === "SECRETARY" && officer.isCorporate === false ) {
+    if (officer.role === OFFICER_ROLE.SECRETARY && officer.isCorporate === false ) {
```
markpit, line 98: can you convert officer.role to capitals and keep the enum as capital, we won't have to worry about matching case then

markpit (approved):

PR #504, opened 2022-01-07
BI-10136 displaying new active officers details page with dummy data

SVarallo:


markpit:
Approving, the duplicate code is in the template which we have stories to split into 'includes' at a later date

markpit (approved):

PR #510, opened 2022-01-11
BI-10140 Removed unused '4 screen' code artifacts

mwestacott:
* Controller, view and supporting code constants and functions removed as now unused


mattch1:
I don't think we need a separate active directors controller anymore it looks like duplicate of the controller you have changed. What do people think?

markpit (approved):

mattch1:
I guess we still need that directors controller for some reason then.

mwestacott:
> I don't think we need a separate active directors controller anymore it looks like duplicate of the controller you have changed. What do people think?

Not sure. I haven't actually changed any controllers in this patchset (just deleted a couple). Are you happy to approve @mattch1 given @markpit 's comment and approval?

PR #512, opened 2022-01-11
BI-10147 extract and display corporate secretary data

SVarallo:


src/controllers/tasks/active.officers.details.controller.ts
```diff
@@ -43,6 +45,24 @@ const buildSecretaryList = (officers: ActiveOfficerDetails[]): any[] => {
     });
 };
 
+const buildCorporateSecretaryList = (officers: ActiveOfficerDetails[]): any[] => {
+  return officers
```
mattch1, line 49: Could rename this to buildCorporateOfficerList as it can be reused for corporate directors. We can see what people think about this.

SVarallo: True it seems the fields are identical so this seems like a good shout 👍

views/tasks/active-officers-details.html
```diff
@@ -158,7 +103,7 @@ <h2 class="govuk-heading-">1 corporate director</h2>
       {% if naturalSecretaryList %}
         {% set officerRows = (officerRows.push([{html: naturalSecretaryHTML}]), officerRows) %}
       {% endif %}
-      {% set officerRows = (officerRows.push([{html: corpSecretaryHTML}]), officerRows) %}
+      {% set officerRows = (officerRows.push([{html: corporateSecretaryHTML}]), officerRows) %}
```
mwestacott, line 106: I think this also needs the check around it to see if there are any rows to actually show.

mattch1 (approved):

PR #517, opened 2022-01-14
BI-10195 Officer identification type lookup

mattch1:
New function in api.enumerations module


src/controllers/tasks/active.officers.details.controller.ts
```diff
@@ -52,7 +53,7 @@ const buildCorporateOfficerList = (officers: ActiveOfficerDetails[], wantedOffic
       return {
         dateOfAppointment: officer.dateOfAppointment,
         forename: formatTitleCase(officer.foreName1),
-        identificationType: officer.identificationType,
+        identificationType: lookupIdentificationType(officer.identificationType as string),
```
markpit, line 56: might be better to use the ternary operator here '?'. Not sure of the behaviour if identificationType is not present (it is optional) and we convert it to a string.
perhaps use identificationType ? lookup... : ""

mattch1: Might not be easy to test for this given the way we test the page at atm. We are just checking the while page for specific field but checking for the expected "" could just lead to false positives.

mwestacott: Agreed, if we are testing based on the page content, but could you possibly mock out the 'lookupIdentificationType' function using Jest in a specific test in which the API returns a null and then check that the mock is NOT called?

test/controllers/tasks/active.officers.enumerations.unit.ts
```diff
@@ -0,0 +1,22 @@
+jest.mock("../../../src/utils/api.enumerations");
```
markpit, line 1: these tests should be in the active.officers.details.controller.unit.ts
If you want to mock a function just for a particular test lifespan you can use 'spy' which gives you teh option to restore the mock to it's original non mocked state

see the task list test as an example - https://github.com/companieshouse/confirmation-statement-web/blob/main/test/controllers/task.list.controller.unit.ts#L137

mattch1: I have tried but it does not seem to work here. I recommend keeping it where it evrything I have tried I get some nasty side effects on the other tests and I really want them all passing.

mwestacott: Is this test actually making a call to the 'buildCorporateOfficerList' function in 'active.officers.details.controller.ts'? Because it's the logic in line 56 of that function that we are trying to exercise/test.

mattch1: Yes we should test this more directly, so well need to export this in a certain way so that we can use spy.
However, even with this it'll still have nasty side effects when put into active.officers.details.controller.unit.ts. It seems to be this line:
jest.mock("../../../src/utils/api.enumerations");

it interfers with the other tests this tells me we should not be mixing them up.

markpit: I'd like to merge this branch in to fix the test - https://github.com/companieshouse/confirmation-statement-web/pull/519
we just have to mock the api enumerations in the controller test and that should fix it.

mattch1: I did think of that but the other tests need to use the actual api enumerations proper; for the corporate officer I am checking that we resolve the chs values to the descriptions. However, I am checing this elsewhere so we could manage without it.

markpit: the api.enumertions has it's own unit test that can check if descriptions get resolved. As the api.enumerations is a dependency of the controller it's ok for us to mock it out.
I think martin's other comment was that the controller wasn't actually being called from the test.

src/controllers/tasks/active.officers.details.controller.ts
```diff
@@ -84,3 +85,7 @@ const buildDirectorList = (officers: ActiveOfficerDetails[]): any[] => {
 const equalsIgnoreCase = (officerRole: string, wantedOfficerRole: OFFICER_ROLE): boolean => {
   return wantedOfficerRole.localeCompare(officerRole, 'en', { sensitivity: 'accent' }) === 0;
 };
+
+export const activeOfficerDetails = {
```
mwestacott, line 89: Not sure this is needed?

mwestacott (approved):

PR #520, opened 2022-01-17
Bi 10197 fix officer data case formatting

markpit:
Adding formatting for the officer details data to convert to title case (upper case first letter, lower case rest of word)
Legal Form values contain some acronyms so have created an exclude list for those and modified teh formatter to take an optional excluded words list.


src/utils/format.ts
```diff
@@ -8,7 +9,7 @@ export const formatTitleCase = (str: string|undefined): string =>  {
 
   return str.replace(
     /\w\S*/g, (word) => {
-      return word.charAt(0).toUpperCase() + word.substr(1).toLowerCase();
+      return word.charAt(0).toUpperCase() + word.substring(1).toLowerCase();
```
markpit, line 12: substr was deprecated, substring is also available as a function on string object

markpit, line 12: substr was deprecated, substring is also available as a function on string object

SVarallo (approved):

PR #521, opened 2022-01-18
BI-10192 adding post method to active officers details controller

SVarallo:


src/controllers/tasks/active.officers.controller.ts
```diff
@@ -66,7 +66,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
       const residentialAddress = formatAddressForDisplay(activeOfficerDetails?.residentialAddress);
       return res.render(Templates.ACTIVE_OFFICERS, {
         backLinkUrl: urlUtils.getUrlToPath(TASK_LIST_PATH, req),
-        officerErrorMsg: DIRECTOR_DETAILS_ERROR,
+        officerErrorMsg: OFFICER_DETAILS_ERROR,
```
markpit, line 69: this controller is for the current live version with 1 director only. Do we know if that error message is supposed to change from director to officer for the current live version?

SVarallo: Admittedly no, shall we just add a new error message for officers then and just leave the directors error as is?

markpit: I would create new message, unless we have had a requirement to change the existing one then keep them separate

markpit (approved):

views/tasks/active-officers-details.html
```diff
@@ -65,8 +71,8 @@ <h1 class="govuk-heading-l">Check the officers' details</h1>
       {{ govukRadios({
         classes: "govuk-radios",
         idPrefix: "officers",
-        name: "officers",
-        errorMessage: directorDetailsError,
+        name: "activeOfficers",
```
mwestacott, line 74: This is fine but also delivers what I was working on (not very much!) - https://companieshouse.atlassian.net/browse/BI-10191. So we can mark that task as done too.

SVarallo: Should still need the data event ids added, I can put that in this as well if that's easier?

PR #529, opened 2022-01-24
BI-10290 Extract the other registrable person data and display it on the active-psc-details page

SVarallo:


src/controllers/tasks/people.with.significant.control.controller.ts
```diff
@@ -55,7 +55,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
       const pscList = formatPscList(pscs);
       return res.render(Templates.PEOPLE_WITH_SIGNIFICANT_CONTROL, {
         backLinkUrl: urlUtils.getUrlToPath(TASK_LIST_PATH, req),
-        pscList: pscList,
+        individualPscList: pscList,
```
markpit, line 58: same here for renaming it back to pscList

markpit (approved):

PR #530, opened 2022-01-25
BI-10291 Redirecting to psc-statement page if company has no PSCs

SVarallo:


src/controllers/tasks/active.psc.details.controller.ts
```diff
@@ -14,13 +14,18 @@ import {
...
 import { toReadableFormat } from "../../utils/date";
+import { logger } from "../../utils/logger";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
     const transactionId = urlUtils.getTransactionIdFromRequestParams(req);
     const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
     const pscs: PersonOfSignificantControl[] = await getPscs(req.session as Session, transactionId, submissionId);
-
+    if (pscs.length < 1) {
```
markpit, line 24: pscs might be undefined as well, i would include that in the if check like
if (!pscs || pscs.length < 1)

test/controllers/tasks/active.psc.details.controller.unit.ts
```diff
@@ -1,14 +1,17 @@
+import { urlUtils } from "../../../src/utils/url";
```
markpit, line 1: move the urlUtils import so it is with the other imports, but make sure that the import mocks is still the first import, safer to add to bottom of imports.

test/controllers/tasks/active.psc.details.controller.unit.ts
```diff
@@ -67,5 +70,16 @@ describe("Active psc details controller tests", () => {
       expect(response.text).toContain("Charity - Unincorporated Association");
       expect(response.text).toContain("Ownership of voting rights - more than 75%");
     });
+
+    it("should navigate to psc statement page if no psc is found", async () => {
```
markpit, line 74: also do a similar test that returns undefined from mockGetPscs.mockResolvedValueOnce
to test that it goes to the psc statement page as well (same as the test when [] gets returned)

test/controllers/tasks/active.psc.details.controller.unit.ts
SVarallo: Ah my bad, it did that automatically

markpit (approved):

PR #532, opened 2022-01-25
BI-10288 adding individual PSC page

SVarallo:


markpit:
also need to revert the changes in normal-psc as the order of data was changed which will be picked up by the phase 1 journey

markpit (approved):

PR #538, opened 2022-01-28
BI-10411 Display correct text on Statement of capital stop screen

mattch1:
Removed else clause as the text is the same in both of the previous conditions.

### DIFF
--- a/views/incorrect-information/wrong-statement-of-capital.html
+++ b/views/incorrect-information/wrong-statement-of-capital.html
@@ -26,14 +26,12 @@
             <p>The company's share capital does not match the shares held by its shareholders.</p>
           {% endif %}
 
-          <p>You must update this information by filing a confirmation statement through
-            <a target="_blank" rel="noopener noreferrer" href="https://ewf.companieshouse.gov.uk//seclogin?tc=1" data-event-id="webfiling-service-link-from-invalid-share-or-amount-unpaid-totals-wrong-statement-of-capital-page">our WebFiling service</a>. You will need to sign in to or create a WebFiling account to do this.</p>
-
-        {% else %}
-          <p>Currently, changes to share capital can only be made by filing a confirmation statement through
-            <a target="_blank" rel="noopener noreferrer" href="https://ewf.companieshouse.gov.uk//seclogin?tc=1" data-event-id="webfiling-service-link-from-wrong-statement-of-capital-page">our WebFiling service (opens in a new tab)</a>. You will need to sign in to or create a WebFiling account to do this.
-          </p>
         {% endif %}
+
+        <p>You must update this information by filing a confirmation statement through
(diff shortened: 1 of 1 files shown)


views/incorrect-information/wrong-statement-of-capital.html
```diff
@@ -26,14 +26,12 @@ <h1 class="govuk-heading-xl">
...
 
-          <p>You must update this information by filing a confirmation statement through
-            <a target="_blank" rel="noopener noreferrer" href="https://ewf.companieshouse.gov.uk//seclogin?tc=1" data-event-id="webfiling-service-link-from-invalid-share-or-amount-unpaid-totals-wrong-statement-of-capital-page">our WebFiling service</a>. You will need to sign in to or create a WebFiling account to do this.</p>
-
-        {% else %}
-          <p>Currently, changes to share capital can only be made by filing a confirmation statement through
-            <a target="_blank" rel="noopener noreferrer" href="https://ewf.companieshouse.gov.uk//seclogin?tc=1" data-event-id="webfiling-service-link-from-wrong-statement-of-capital-page">our WebFiling service (opens in a new tab)</a>. You will need to sign in to or create a WebFiling account to do this.
-          </p>
         {% endif %}
+
```
markpit, line 30: Looks like there are 2 different matomo events in the old else block urls depending on what branch of the if statement gets executed so we'll need to keep those unless the story says to remove them

views/incorrect-information/wrong-statement-of-capital.html
```diff
@@ -26,14 +26,12 @@ <h1 class="govuk-heading-xl">
...
+        <p>You must update this information by filing a confirmation statement through
+          <a href="https://ewf.companieshouse.gov.uk//seclogin?tc=1">our WebFiling service</a>. You will need to sign in to or create a WebFiling account to do this.
```
markpit, line 32: the old links used to open in a new tab, so we should do the same with this new url unless the story tells us not to.

views/incorrect-information/wrong-statement-of-capital.html
markpit, line 30: line 19 & 29
{% if totalAmountUnpaidValidation === false or sharesValidation === false %}
is not needed anymore as it's contents each have their own if.

mattch1: Are you sure? It will clash with you comment above about the matomo events.

markpit: just noticed it in the PR as it stands now.
currently it is 
if (a or b) {
 if (a) {..}
 if (b) {..}
}

so you can just do 
if(a) {..}
if (b) {..}

might want to keep it if you are adding the else block back, depends how you want to build the url

markpit (approved):

PR #540, opened 2022-01-31
BI-10414 display the correct number of orps

mattch1:


mwestacott (commented):
Just wondering if you could have written a test to replicate the defect (before fixing) which would then prevent any regression in the future?

mwestacott (approved):

PR #545, opened 2022-02-03
Bi 10494 new controller invalid company status

markpit:
Added new controller for invalid.company.status stop page

Modified the existing confirm.company.controller to create a new map for mapping eligibility codes to redirect paths (need to move from res.render template to redirect to a controller)

### DIFF
--- a/src/types/page.urls.ts
+++ b/src/types/page.urls.ts
@@ -7,6 +7,7 @@ export enum urlParams {
 }
 
 export enum URL_QUERY_PARAM {
+  COMPANY_NUMBER = "companyNumber",
   IS_PSC = "isPsc"
 }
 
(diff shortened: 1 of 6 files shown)


test/controllers/invalid.company.status.controller.unit.ts
```diff
@@ -0,0 +1,49 @@
...
+describe("Invalid company status controller tests", () => {
+
+  beforeEach(() => {
+    mocks.mockAuthenticationMiddleware.mockClear();
+    mocks.mockServiceAvailabilityMiddleware.mockClear();
+    mocks.mockSessionMiddleware.mockClear();
+    mockGetCompanyProfile.mockClear();
+  });
+
+  describe("get tests", () => {
```
mattch1, line 25: Could we rename this so it's a bit clearer to "get function tests" or "test get function"?

src/types/page.urls.ts
```diff
@@ -7,6 +7,7 @@ export enum urlParams {
 }
 
 export enum URL_QUERY_PARAM {
+  COMPANY_NUM = "companyNumber",
```
mattch1, line 10: I'd name this COMAPNY_NUMBER

markpit: I did try that but Sonar raised it as a code smell due to there being another COMPANY_NUMBER already in the file so had to come up with another name in keeping with the other Enum IS_PSC

mattch1 (approved):

SVarallo:
Can we check the matomo console tracks the new url?

SVarallo (approved):
Confirmed it all shows up as intended on matomo

PR #548, opened 2022-02-07
BI-10494 Add company number query param validation in invalid.company…

markpit:
….status.controller

Adding a call to validate the company number that is used in the query param to prevent any other user data being added.
If it is not a valid company number, service throws error and stops the user.

### DIFF
--- a/src/controllers/invalid.company.status.controller.ts
+++ b/src/controllers/invalid.company.status.controller.ts
@@ -2,7 +2,7 @@ import { NextFunction, Request, Response } from "express";
 import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
 import { getCompanyProfile } from "../services/company.profile.service";
 import { Templates } from "../types/template.paths";
-import { URL_QUERY_PARAM } from "../types/page.urls";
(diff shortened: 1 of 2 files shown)


mattch1 (approved):

src/controllers/invalid.company.status.controller.ts
```diff
@@ -3,10 +3,16 @@ import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/compa
...
 import { URL_QUERY_PARAM } from "../types/page.urls";
+import { isCompanyNumberValid } from "../validators/company.number.validator";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
-    const companyNumber = req.query[URL_QUERY_PARAM.COMPANY_NUM];
+    const companyNumber = req.query[URL_QUERY_PARAM.COMPANY_NUM] as string;
+
+    if (!isCompanyNumberValid(companyNumber)) {
+      return next(new Error("Invalid company number entered in invalid.company.status url query parameter"));
```
mwestacott, line 13: Rather than reference 'invalid.company.status' here could you maybe use the INVALID_COMPANY_STATUS_PATH variable/constant? Just a thought.

mwestacott (approved):

PR #549, opened 2022-02-07
Bi 10495 change url file on paper stop page

mattch1:
Controller and corresponding routing and unit tests for the sue paper stop page.
Updated the confirm.company.controller unit tests to reflect the change.

As we are using query parameters, have added a condition to catch invalid company numbers.


src/controllers/paper.filing.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { URL_QUERY_PARAM } from "../types/page.urls";
+import { CompanyProfile } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
+import { getCompanyProfile } from "../services/company.profile.service";
+import { isCompanyNumberValid } from "../validators/company.number.validator";
+
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const companyNumber = req.query[URL_QUERY_PARAM.COMPANY_NUM] as string;
+    if (!isCompanyNumberValid(companyNumber)) {
+      return next(new Error("Invalid company number entered in invalid.company.status url query parameter"));
```
markpit, line 12: the error has the incorrect controller name in the message. Could do as martin suggest on my PR and replace the controller name with the paper filing path (but don't substitute the company number as we shouldn't log user data incase it is malicious causing log injection attack)

src/controllers/paper.filing.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+    }
+    const company: CompanyProfile = await getCompanyProfile(companyNumber);
+    return res.render(Templates.USE_PAPER, {
+      company,
+      templateName: Templates.USE_PAPER
+    });
+  } catch (e) {
+    return next(e);
```
markpit, line 20: line 20 needs a unit test to verify error page returned when exception is caught

markpit (approved):

PR #553, opened 2022-02-08
BI-10485 Use a controller to redirect to the tradings top page

mattch1:
Navigate to the trading stop page as we do to the task list page, using url values for the backlink.


src/controllers/trading.stop.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Templates } from "../types/template.paths";
+import { urlUtils } from "../utils/url";
+import { TRADING_STATUS_PATH } from "../types/page.urls";
+
+export const get = (req: Request, res: Response) => {
+
+  const companyNumber = urlUtils.getCompanyNumberFromRequestParams(req);
+  const transactionId = urlUtils.getTransactionIdFromRequestParams(req);
+  const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
+  return res.render(Templates.TRADING_STOP, {
```
markpit, line 11: there's a more concise way to get the url - you can use this
return res.render(Templates.TRADING_STOP, {
      backLinkUrl: urlUtils.getUrlToPath(TRADING_STATUS_PATH, req)
    });

test/controllers/trading.stop.controller.unit.ts
```diff
@@ -0,0 +1,31 @@
+jest.mock("../../src/validators/company.number.validator");
+
+import mocks from "../mocks/all.middleware.mock";
+import { TRADING_STOP_PATH } from "../../src/types/page.urls";
+import request from "supertest";
+import app from "../../src/app";
+import { isCompanyNumberValid } from "../../src/validators/company.number.validator";
```
markpit, line 7: don't need the isCompanyNumberValid mock in this test

markpit (approved):

PR #556, opened 2022-02-09
BI-10486 Use controller to display the wrong sic stop page

mattch1:
Reroute to a controller so that the wrong sic page has its own url.


test/controllers/incorrect-information/wrong.sic.controller.unit.ts
```diff
@@ -0,0 +1,29 @@
...
+  describe("test for the get function", () => {
+
+    it("Should render the wrong sic stop page", async () => {
+      const response = await request(app).get(WRONG_SIC_PATH);
+
+      expect(response.text).toContain(STOP_PAGE_TEXT);
+      expect(response.text).not.toContain(SIC_CODE_DETAILS);
+      expect(response.text).toContain("Incorrect SIC");
+    });
+  });
```
markpit, line 28: I'd add a test to check the back link url is correct as well

mattch1: Not sure whether there's any easy way to do this and make a test meaningful, do we usuually check for backlinks?

markpit: i think we've done it in a couple of the tests. I'll see if I can find a way.

mattch1: I have added this as long as we just need to prove that it links back to the sic page then it should be fine.

markpit (approved):

PR #563, opened 2022-02-14
BI-10492 Redirect to stop page with controller

mattch1:
Need to separate out the controller and stop page urls, but they still use the same template


src/routes/routes.ts
```diff
@@ -28,7 +28,8 @@ import * as wrongStatementOfCapitalRoute from "../controllers/incorrect-informat
 import * as wrongShareholdersRoute from "../controllers/incorrect-information/wrong.shareholders.controller";
 import * as wrongRegisteredOfficeAddressRoute from "../controllers/incorrect-information/wrong.registered.office.address.controller";
 import * as wrongRegisterLocationsRoute from "../controllers/incorrect-information/wrong.registers.controller";
-import * as wrongDetailsRoute from "../controllers/incorrect-information/wrong.details.controller";
+import * as wrongDetailsRoute from "../controllers/incorrect-information/wrong.officer.details.controller";
```
mwestacott, line 31: Could you import this with the name 'wrongOfficerDetailsRoute' which makes clear what it is? Consistent then with 'wrongPscDetailsRoute'.

src/types/page.urls.ts
```diff
@@ -79,5 +79,7 @@ export const WRONG_RO =  ACTIVE_SUBMISSION_BASE + Templates.WRONG_RO;
 export const WRONG_RO_PATH = CONFIRMATION_STATEMENT + WRONG_RO;
 export const WRONG_REGISTER_LOCATIONS = ACTIVE_SUBMISSION_BASE + Templates.WRONG_REGISTER_LOCATIONS;
 export const WRONG_REGISTER_LOCATIONS_PATH = CONFIRMATION_STATEMENT + WRONG_REGISTER_LOCATIONS;
-export const WRONG_DETAILS = ACTIVE_SUBMISSION_BASE + Templates.WRONG_DETAILS;
+export const WRONG_DETAILS = ACTIVE_SUBMISSION_BASE + "wrong-officer-details";
```
mwestacott, line 82: So would then suggest refactoring/renaming this to WRONG_OFFICER_DETAILS too.

mattch1: The template is used by both the officers and psc page so needs to stay as generic as possible.

mattch1: Ignore my reply changing this.

mwestacott (approved):

PR #564, opened 2022-02-15
BI-10493 Display wrong psc statement page using a controller

mattch1:
As with psc details added an independent controller and url.


src/controllers/incorrect-information/wrong.officer.details.controller.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export const get = (req: Request, res: Response) => {
-  return res.render(Templates.WRONG_DETAILS, {
-    templateName: Templates.WRONG_DETAILS,
-    backLinkUrl: urlUtils.getUrlToPath(ACTIVE_OFFICERS_DETAILS_PATH, req),
-    returnToTaskListUrl: urlUtils.getUrlToPath(TASK_LIST_PATH, req),
-    stepOneHeading: WRONG_DETAILS_UPDATE_OFFICER,
-    pageHeading: WRONG_DETAILS_UPDATE_OFFICERS,
-  });
+  return res.render(Templates.WRONG_DETAILS, buildWrongDetails(
```
markpit, line 11: i don't think the buildWrongDetails makes much of an improvement. I prefer it as it was as we can see what fields are being sent to the template without having to click out of the controller into the buildWrongDetails. See what @mwestacott thinks.

mattch1: Sonar was complaining about duplication.

src/services/wrong.details.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Request } from "express";
+import { Templates } from "../types/template.paths";
+import { urlUtils } from "../utils/url";
+
+export const buildWrongDetails = (req: Request, backLink: string, returnPath: string,
+                                  stepOneHeading: string, pageHeading: string) => {
+  return {
```
markpit, line 7: if we keep this approach then the service will need it's own unit test.

mattch1: Its is covered by the existing controller tests so anything changed here would be picked up by those tests.

src/controllers/incorrect-information/wrong.officer.details.controller.ts
mattch1: Removed new service and put the controllers back to how they were.

mwestacott (approved):

markpit (approved):

PR #565, opened 2022-02-18
BI-10598 adding radio button id validator and adding it to controller classes

SVarallo:


src/controllers/tasks/active.officers.controller.ts
```diff
@@ -45,6 +46,10 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
     const submissionId = urlUtils.getSubmissionIdFromRequestParams(req);
     const session: Session = req.session as Session;
     const activeOfficerDetailsBtnValue = req.body.activeOfficers;
+
+    if (!isRadioButtonValid(activeOfficerDetailsBtnValue)) {
+      return next(new Error("No valid radio button id in request"));
```
mwestacott, line 51: See my other comment but I think the log output here and below needs to change slightly as it's the radio button value/choice which is invalid and not the id of the radio button.

src/validators/radio.button.validator.ts
```diff
@@ -0,0 +1,24 @@
...
+export const isRadioButtonValid = (radioId: string): boolean => {
+
+  logger.debug("Check radio button is valid");
+
+  if (!radioId) {
+    logger.debug("No radio button id supplied");
+    return true;
+  }
+
+  const validRadioIds: string[] = ["yes", "no", "recently_filed"];
```
markpit, line 12: you should use the RADIO_BUTTON_VALUES enum from utils/constants instead of the strings 'yes' 'no' 'recently_filed'

test/validators/radio.button.validator.unit.ts
```diff
@@ -0,0 +1,27 @@
+import { isRadioButtonValueValid } from "../../src/validators/radio.button.validator";
+
+describe("radio button validator tests", () => {
+
+  describe("isRadioButtonValid tests", () => {
```
markpit, line 5: could do with a test for the getRadioButtonInvalidValueErrorMessage function

markpit (approved):

PR #567, opened 2022-02-28
Bi 10598 split id validation middleware

markpit:
Split the validation middleware up into 2 so they are independent (could have url with either or both txn and submission ids)

Checks if ids are less than 50 chars in length and will show the 'service unavailable' screen if invalid with a 400 http response

As the url can contain a few ids that can be upto 50 chars each, i've truncated the whole url to 400 before logging it out in the case of validation failure.

The CH logger library would automatically include the url in the log statement so I've added function to truncate it before logging which is used on validation failure.

Logging validation failures at INFO level rather than ERROR


src/middleware/transaction.id.validation.middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+  logger.debug("Execute URL transaction id validation middleware checks");
+
+  const transactionId: string = req.params[urlParams.PARAM_TRANSACTION_ID];
+
+  logger.debug("Check transaction id");
+  if (!isUrlIdValid(transactionId)) {
+    // need to truncate the url in the request as we are about to log it out and it has failed validation.
+    // it could contain a large amount of data so truncate to stop logs filling up
+    urlUtils.truncateRequestUrl(req);
+    logger.infoRequest(req, "No Valid Transaction Id in URL: " + req.originalUrl.substring(0, URL_LOG_LENGTH));
```
mwestacott, line 19: As discussed, suggest using 'error' level logging here as no 'warning' level available and this is closer to an error than just information. If support does have some tool monitoring the logs in LIVE for errors then it also means this will be flagged up.

src/middleware/transaction.id.validation.middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+export const transactionIdValidationMiddleware = (req: Request, res: Response, next: NextFunction) => {
+  logger.debug("Execute URL transaction id validation middleware checks");
+
+  const transactionId: string = req.params[urlParams.PARAM_TRANSACTION_ID];
+
+  logger.debug("Check transaction id");
+  if (!isUrlIdValid(transactionId)) {
+    // need to truncate the url in the request as we are about to log it out and it has failed validation.
+    // it could contain a large amount of data so truncate to stop logs filling up
+    urlUtils.truncateRequestUrl(req);
```
mwestacott, line 18: Rather than truncate the entire URL, suggest replacing and truncating just the id which is too long. Could append elipses (...) on to the end like in the id validator.

mwestacott (approved):

PR #569, opened 2022-03-03
Bi 10707 add ispsc query param validation

mwestacott:


src/middleware/is.psc.validation.middleware.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const isPsc: string = req.query[URL_QUERY_PARAM.IS_PSC] as string;
+
+  if (!isPsc) {
+    return next();
+  }
+
+  logger.debug("Check isPsc");
+  if (!isPscFlagValid(isPsc)) {
+    logger.errorRequest(req, "No valid isPsc query parameter supplied: " + req.originalUrl);
```
mwestacott, line 18: @markpit I think I'll need to make use of your new common function to truncate the request content at this point, once merged.

src/middleware/is.psc.validation.middleware.ts
```diff
@@ -0,0 +1,23 @@
...
+import { isPscFlagValid } from "../validators/is.psc.validator";
+import { logger } from "../utils/logger";
+import { Templates } from "../types/template.paths";
+
+export const isPscQueryParameterValidationMiddleware = (req: Request, res: Response, next: NextFunction) => {
+  logger.debug("Execute isPsc query parameter validation middleware checks");
+
+  const isPsc: string = req.query[URL_QUERY_PARAM.IS_PSC] as string;
+
+  if (!isPsc) {
```
mattch1, line 12: Do we need this check in two places?

src/validators/is.psc.validator.ts
```diff
@@ -0,0 +1,12 @@
+export const isPscFlagValid = (isPsc: string): boolean => {
+
+  if (!isPsc) {
```
mattch1, line 3: Do we want to check for false specifically rather than for the more generic falsy? (as that would cover undefined as well)

src/middleware/is.psc.validation.middleware.ts
markpit: the query param will usually not be present and will be undefined. Trouble is we can't filter the url by a query param so all urls will have to enter this middleware and the majority won't have this param and it will be undefined. If we let it through to validation then it will fail (as undefined is not an allowed isPsc value) and the user will be blocked.

src/validators/is.psc.validator.ts
markpit: 'false' is a string and won't cover undefined. False is a valid value, the accepted isPsc values can be 'true' or 'false'.

markpit: i think i see your point, the second check should return false if undefined so the first check might be redundant

src/middleware/is.psc.validation.middleware.ts
mwestacott: Worth adding a comment in the code explaining this. I'll do that.

src/validators/is.psc.validator.ts
mattch1: We could keep the first check for the reason you pointed out above and then here just check for true and return true and then false in all other cases, so
if (isPsc) { return true; } return false;

mwestacott: Ok. Will investigate if this first check can be removed.

mwestacott: > We could keep the first check for the reason you pointed out above and then here just check for true and return true and then false in all other cases, so if (isPsc === "true") { return true; } return false;

We can't just return true when 'isPsc' is 'true' as an 'isPsc' value of 'false' is also valid.

I'm removing that first check though, as it's not needed.

test/middleware/is.psc.validation.middleware.unit.ts
```diff
@@ -0,0 +1,51 @@
...
+mockIsPscValidator.mockReturnValue(true);
+
+describe("is PSC validation middleware tests", () => {
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should call next() when isPsc query parameter not present", () => {
+    req.query = { isPsc: undefined };
```
markpit, line 26: i's also add a test to replicate no query string present (might have to console.log the req.query value in the middleware to see what it looks like when no query string is in url).
so have a test where req.query = {}  ( or req.query = undefined, or both!)

mwestacott: One new test added for req.query = {}. req.query = undefined causes an error and doesn't represent what comes through when no query params are present.

src/middleware/is.psc.validation.middleware.ts
mwestacott: This now added and will be present in next commit.

markpit (approved):

PR #571, opened 2022-03-03
BI-10757 Externalising the validation constants to config file

markpit:
Moving the validation constants to environment props aka config file.
Set default values as per the previous constants in case they are missing in config file


mwestacott (approved):
Will there be a docker config change to add these config values too? I know the defaults will apply but perhaps useful to have all configurable params present in the docker config?

markpit:
> Will there be a docker config change to add these config values too? I know the defaults will apply but perhaps useful to have all configurable params present in the docker config?

yes, didn't have time last night so will get those done next, I won't merge until config in

PR #577, opened 2022-03-09
BI-10602 adding encodeUrls function to url utils and unit tests

SVarallo:


src/utils/url.ts
```diff
@@ -91,6 +99,6 @@ export const urlUtils = {
   getUrlToPath,
   getUrlWithCompanyNumber,
   getUrlWithCompanyNumberTransactionIdAndSubmissionId,
-  sanitiseReqlUrls,
+  sanitiseReqlUrls: sanitiseReqUrls,
```
mwestacott, line 102: Rather than do this and keep supporting the misspelt function name I would just have sanitiseReUrls and update all the places where the old function name was used.

mwestacott (approved):

PR #589, opened 2022-05-17
Correction to the wrong register stop screen.

rgoswami-ch:
Correction to the text for wrong register stop screen to match the prototype.


DTaylor-CH (approved):
LGTM

PR #594, opened 2022-05-26
Fixing issues with stop screen radio button alerts

rgoswami-ch:
Moved condition check away from alert fragment to main html.

[BI-11032](https://companieshouse.atlassian.net/browse/BI-11032)


rmcmurray-ch (approved):
LGTM 👍

PR #598, opened 2022-06-01
Fix the back link from the PSC Statement stop screen

rgoswami-ch:
[BI-11151](https://companieshouse.atlassian.net/browse/BI-11151)

Added a back button method in the psc statement stop screen controller which generates the back button based on whether there are PSCs present or not.


rmcmurray-ch (approved):
LGTM - back link working as expected now when tested locally, code looks good 👍

PR #615, opened 2022-07-14
Updating the stop screens for matomo tracking

rgoswami-ch:
Splitting up the wrong-detail.html into individual templates for the stop screens that use the template (officers, psc details, psc statement)


src/types/template.paths.ts
```diff
@@ -22,6 +22,8 @@ export enum Templates {
   TRADING_STOP = "trading-stop",
   NO_FILING_REQUIRED = "no-filing-required",
   WRONG_DETAILS = "incorrect-information/wrong-details",
```
sbasile-ch, line 24: Can `WRONG_DETAILS` be removed now ?

PR #621, opened 2022-08-01
Signout - Add signout to confirmation screen

HarryBosworth:
Added signout bar to confirmation screen. 
Added configuration to signout bar to allow signout url to change.
In the confirmation controller, set the signoutURL so that the signout cofirmation screen is by passed. 

Addition to [BI-11289](https://companieshouse.atlassian.net/browse/BI-11289)


rmcmurray-ch (approved):
LGTM 👍

PR #622, opened 2022-08-09
Added motomo tracking events to signout links

HarryBosworth:
- Emit 'signout' event when user clicks 'signout' in the signout bar
- Emit 'signout-yes' event when the user selects 'yes' on the signout screen
- Emit 'signout-no' event when the user selects 'no' on the signout screen


sbasile-ch (approved):
ok

PR #651, opened 2023-08-16
Change LOGLEVEL to LOG_LEVEL

bcullerton:
Match to level set in the structured logger config:

level: process.env.LOG_LEVEL === undefined ? "info" : process.env.LOG_LEVEL.toLowerCase(),


tobafolaranmi-ch (approved):
LGTM

PR #674, opened 2023-10-26
Check existing email address if it exists [ECCT-762]

nevans4-ch:
Clarified naming - renamed existing registered-email-address to provide-email-address & CHECK items to CONFIRM to keep a clear distinction between check-, provide- & confirm- screens and related code.

Added check-email-address page & updated task list page to use this as target url for registered email address task IF one already exists.

Used new endpoint in private SDK to check for existing email address:

- added private-api-sdk-node as a dependency
- tweak to Dockerfile to enable it to get this from git
- added call to cs-api via private-api-sdk-note to get email address
 & added tests


adeharrison-ch (changes requested):

adeharrison-ch (approved):
LGTM

PR #703, opened 2023-12-12
add two new checkboxes with accompanying html text, centralise common ecct function

CHssmith2:
### JIRA link

https://companieshouse.atlassian.net/browse/ECCT-616

### Change description

1. Add two new checkboxes for the following statements:
     Confirmation Statement Text
     Lawful Activity Statement Text

2. Centralise existing ECCT Day One code so it can be used in multiple places.
3. Added new unit tests for the centralised ecct code as well as add new unit tests for the updated review controller and web html page.

### DIFF
--- a/views/review.html
+++ b/views/review.html
@@ -1,4 +1,5 @@
 {% extends "layout.html" %}
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
 
 {% block pageTitle %}
     Submit the confirmation statement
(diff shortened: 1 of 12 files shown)


views/review.html
```diff
@@ -1,4 +1,5 @@
 {% extends "layout.html" %}
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
```
nevans4-ch, line 2: Does this belong in layout.html with other imports?

nevans4-ch (changes requested):
Looks good - just one nitpick about an import placement, comment added inline.

views/review.html
CHssmith2: Good spot - now moved to the layout.html page.

nevans4-ch (approved):

PR #742, opened 2024-08-15
Update paper fee amount & paper form url [BI-13955]

nevans4-ch:
Update paper fee amount from £40 to £62 (in service unavailable page as well as start page).
Also update all instances of paper form link (start page, service unavailable page, paper filing page) to point to post-ECCT version.


tcbaileych (approved):
approved, looks good to me

PR #793, opened 2024-10-18
Sec 112 enable add csrf middleware

mattch1:
https://companieshouse.atlassian.net/browse/SEC-112

Added middleware to the app.
Updated error controller and added a unit test.
Added html error page template.

Tested unit tests with enabled true and false.
Tested journey with enabled false.


mwestacott (approved):
LGTM but I would get another review before merging.

dlloyd1-ch (approved):
LGTM

PR #863, opened 2025-08-20
CSE-838: Allowing subtype to be used within company profile 

sstevenson-ch:
New "subtype" added to company profile 

Tests added and updated
Integrated to work with feature flags

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


package-lock.json
imclaughlin-ch, line 1: Can you remove this from the commit please. The lock file is generated automatically when installing new dependencies

imclaughlin-ch (changes requested):

dmcfaull-ch (approved):

dho-ch (approved):

PR #869, opened 2025-08-28
Cse 880 review controller refactor

dmcfaull-ch:
Refactor review controller to reduce complexity


src/utils/limited.partnership.ts
```diff
@@ -85,3 +91,59 @@ export function isLimitedPartnershipSubtypeFeatureFlagEnabled(companyProfile: Co
   }
   return false;
 }
+
+export function handleLimitedPartnershipConfirmationJourney (req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, res: Response<any, Record<string, any>>, companyNumber: string, companyProfile: CompanyProfile, transactionId: string, submissionId: string, session: Session) {
```
imclaughlin-ch, line 95: Can we extract this logic into a more specific util for the review controller for limited partnership handling?

So something like this package structure:

utils/confirmation/limited.partnership.confirmation.ts

src/utils/no.change.ts
imclaughlin-ch, line 1: Similar comment on the package structure:

Can we extract this logic into a more specific util for the review controller for no change journey handling?

So something like this package structure:

utils/confirmation/no.change.confirmation.ts

imclaughlin-ch (changes requested):

sstevenson-ch (approved):

kensayle-ch (approved):

PR #967, opened 2026-03-26
Enable vscode docker debugger functionality

AlisonGriffiths:


nodemon.json
```diff
@@ -1,5 +1,5 @@
 {
-  "exec": "ts-node ./src/bin/nodemon-entry.ts",
+   "exec": "node --inspect=0.0.0.0:48132 -r ts-node/register src/bin/nodemon-entry.ts",
   "ext": "ts,html",
   "watch": ["./src", "./views"],
```
PhilipGlover: I have, but whether all are needed I'm not sure..
```
"ext": "ts,html,njk,json,js",
"watch": ["./src", "./locales", "./views"],
```

nodemon.json
```diff
@@ -1,5 +1,5 @@
 {
-  "exec": "ts-node ./src/bin/nodemon-entry.ts",
+   "exec": "node --inspect=0.0.0.0:48132 -r ts-node/register src/bin/nodemon-entry.ts",
```
PhilipGlover: - extra space at start of `exec` line
- ought this prefaced with dot slash `./src/bin/...`

nodemon.json
PhilipGlover: added

nodemon.json
PhilipGlover: changed as above

PhilipGlover (approved):
matches what done elsewhere, have used

PR #982, opened 2026-04-24
CSE-674: Implement Matomo data events into Check your Answer page

sstevenson-ch:
This pull request introduces enhancements to the Limited Partnership "Check Your Answer" flow, primarily to support improved analytics tracking and template management. The main changes involve adding a new Matomo page name constant, updating the controller to use it, and adding data attributes for event tracking in the Nunjucks template.

**Analytics and Template Improvements:**

* Added a new constant `LP_CS_DATE_CYA` to `MATOMO_LIMITED_PARTNERSHIP_PAGE_NAME` in `constants.ts` to represent the "Check Your Answer" page for confirmation statement date.
* Updated the `get` controller in `lp.check.your.answer.controller.ts` to pass the new `templateName` to the view, enabling correct analytics tracking and template selection.

**Event Tracking Enhancements:**

* Modified the "Change" link for the confirmation statement date in `lp-check-your-answer.njk` to include a `data-event-id="change-cs-date"` attribute for analytics tracking.
* Updated the "Continue" button in the same template to also include a `data-event-id="change-cs-date"` attribute, supporting consistent event tracking.

### DIFF
--- a/views/limited-partners/components/check-your-answer/lp-check-your-answer.njk
+++ b/views/limited-partners/components/check-your-answer/lp-check-your-answer.njk
@@ -31,7 +31,11 @@
                                             {
                                                 href: csDatePageUrl,
                                                 text: i18n.CDCYACheckAnswersChange,
-                                                visuallyHiddenText: i18n.CDCYACheckAnswersChange + " " + i18n.CDCYACheckAnswersStart
+                                                visuallyHiddenText: i18n.CDCYACheckAnswersChange + " " + i18n.CDCYACheckAnswersStart,              
+                                                attributes: {
+                                                        "data-event-id": "change-cs-date"
+                                                    }
+
                                             }    
                                     ]
                                 }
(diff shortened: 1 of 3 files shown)


views/limited-partners/components/check-your-answer/lp-check-your-answer.njk
```diff
@@ -31,7 +31,11 @@
                                             {
                                                 href: csDatePageUrl,
                                                 text: i18n.CDCYACheckAnswersChange,
-                                                visuallyHiddenText: i18n.CDCYACheckAnswersChange + " " + i18n.CDCYACheckAnswersStart
+                                                visuallyHiddenText: i18n.CDCYACheckAnswersChange + " " + i18n.CDCYACheckAnswersStart,              
+                                                attributes: {
+                                                        "data-event-id": "change-cs-date"
```
kensayle-ch: The data-event-id needs to reflect the control it is attached to.

Not entirely sure what it would be in this case but for the continue button something like `continue-button` as the event in Matomo will be prefixed with what ever was passed from the Controller as the `templateName`

kensayle-ch (approved):

PR #986, opened 2026-05-08
ASM-2142 Integrate Open Telemetry into confirmation-statement-web

garethpowell-ch:
Other improvements made:
- Updated dependencies
- Migrated to updated ESLint
- Updated Jest config (but not able to upgrade to latest due to breaking changes)


src/bin/nodemon-entry.ts
```diff
@@ -1,3 +1,4 @@
+import "../openTelemetry"
```
JAndrewCH, line 1: Is this needed?

guyjackson-ch (approved):

src/bin/nodemon-entry.ts
garethpowell-ch: That's what enables Otel in our docker-chs-development environment

PR #1031, opened 2026-07-09
CSE-1329 : Fixing back button bug 

sstevenson-ch:
This pull request updates the `lp.must.be.authorised.agent` controller to improve navigation and localization support for the "must be authorised agent" page. The main change is the addition of a `previousPageWithoutLang` URL to the template context, ensuring users can navigate back to the correct page without language parameters. There is also a minor update to the `Templates` enum formatting.

**Controller improvements:**

* Added construction of a `prevPageURL` (without language parameter) using `urls.CONFIRM_COMPANY_PATH` and the company number extracted from the request, and passed this as `previousPageWithoutLang` to the template context in `lp.must.be.authorised.agent.controller.ts`.

**Code quality/formatting:**

* Added a trailing comma to the `Templates` enum in `template.paths.ts` for consistency.


dmcfaull-ch (approved):
All looks good, being pedantic would it be worth adding a test case?

